PR #2, opened 2021-04-22
SFR-1060: Settings Card Components

siberry:
This PR adds `Toggle`, `ToggleButton` components and a `ToggleGroup` story that reflects the options for the web-reader settings card. At this point, it's mostly CSS changes, either through the Chakra theme or whatever approach we decide on for CSS. 

Currently, `Toggle` accepts an array of either objects (with `value` and `colorScheme for now) or strings when just the value is needed. This could be refactored to use composition.

Off the top of my head, styling needed: 
- the check icon should be `position: absolute` to prevent the button from expanding when checked
- toggles will expand to width of container


kristojorg:
Ok this looks great! I deployed the storybook too so we can check it out on every PR, and I made a couple commits to fix the failing lints on windows (had nothing to do with your changes, just hadn't gotten it working previously).

Some notes

- The button I think needs a border. I guess we can check the other parts of the app to see if it is used without a border elsewhere, but at least in the settings card it has a border. We can also wait for Marcus' design guidance and/or wait to see it in a full card layout before adding one. 
- I really like the toggle! Looks good. One thing I would change is the colors in "active" state (when you're clicking it), because they're currently blueish. I think they should be greenish like the selected state. 
- I agree we should fix how it expands with a checkmark, although I am not sure how exactly. Using position absolute will mean that the button contents are off center if the icon is present. Alternatively, we could always pass an icon, and just make it visible when `isChecked` is true. But then the contents would be off center when there is _no_ icon visible... so maybe the best solution is to always have something be visible? Like an empty checkbox? I think we should wait for design guidance on this.
- I think we should name the `Toggle` to `ToggleGroup` and what is now `ToggleGroup` to `SettingsCard` maybe. When I hear `ToggleGroup` I think of the series of buttons that make up a toggle.
- We should add a few snapshot tests, and maybe a test of the toggling as well.


Design checks we need to do:
- Is this accessible? 
- Are the disabled versions appropriately accessible?
- How should we mark the "selected" button? Previously it had a checkmark, and was _sometimes_ green. You can see the problem with having the checkmark in the storybook on this PR. It makes the button grow when it is present.
- How should the card/modal itself be presented on different screen sizes? It seems to me it would be best to make it full screen on mobile.

Super happy with this PR, it looks great : ). I'll go through the actual code in a little while

src/ui/Toggle.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { useRadioGroup } from '@chakra-ui/react';
+
+import ToggleButton from './ToggleButton';
+
+type optionProps = {
```
kristojorg: Type names should be capitalized (unless they're one of the basic data types like `string` or `number`). Maybe I should add an eslint rule about it come to think of it

src/ui/Toggle.tsx
```diff
@@ -0,0 +1,48 @@
...
+        const radio = getRadioProps({ value });
+        return (
+          <ToggleButton key={value} optionProps={rest} {...radio}>
+            {value}
+          </ToggleButton>
+        );
+      })}
+    </div>
+  );
+};
```
kristojorg: So I know this looks a lot like the custom example that Chakra gave, but I think we want to change up this api a little bit to be more like their standard [Radio](https://chakra-ui.com/docs/form/radio#usage), where instead of having one component that you pass everything into, you use composition. The difference is between these two:

```tsx
const Settings = () => {
  return (
    <ToggleButton
      options={[{value: "blah", colorScheme: "blue"}, {value: "another", colorSheme: "red"}]}
      defaultValue="blah"
    />
  );
}
```
and this:

```tsx

const Settings = () => {
  return (
    <ToggleGroup defaultValue="blah">
      <ToggleButton value="blah" colorScheme="blue" />
      <ToggleButton value="another" colorScheme="red" />
    </ToggleGroup>
  );
}
```

This way, we don't have to pass the props for the `ToggleButton` down through the `Toggle`. You can imaging how that would get complicated if we started wanting to put other props on the `ToggleButton`, like maybe just adding some `className` or a specific style, or making `ToggleButton` support a `size` prop.

src/ui/ToggleButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const input = getInputProps();
+  const checkbox = getCheckboxProps();
+
+  return (
+    <ChakraBox as="label">
+      <input {...input} />
+      <Button as="div" isChecked={isChecked} {...checkbox} {...optionProps}>
+        {children}
+        {isChecked ? <Icon decorative name={IconNames.check} /> : null}
+      </Button>
```
kristojorg: I think you can pass the icon in to the `Button` as `rightIcon`:

```
<Button as="div" rightIcon={isChecked ? <Icon /> : null} />
```

but we still have to figure out how to make it so the button doesn't grow/shrink like crazy

src/ui/theme/components/button.ts
```diff
@@ -1,21 +1,86 @@
...
+    isChecked ? 'green.500' : uncheckedColor;
+  const toggleButtonColor = (uncheckedColor: string) =>
+    isChecked ? 'white' : uncheckedColor;
+
+  switch (colorScheme) {
+    case 'sepia':
+      return {
+        ...common,
+        bg: toggleButtonBg('rgb(246, 236, 217)'),
+        color: toggleButtonColor('rgb(17, 17, 17)'),
```
kristojorg: I have a PR coming after this one that includes all the NYPL colors, so we should be able to change these hard-coded strings to values from the theme like `"ui.success.primary` for the green and `ui.gray.light` for gray etc.

src/ui/theme/components/button.ts
```diff
@@ -1,21 +1,86 @@
...
+  const common = {
+    _focus: {
+      'z-index': 1,
+    },
+  };
+
+  const toggleButtonBg = (uncheckedColor: string) =>
+    isChecked ? 'green.500' : uncheckedColor;
+  const toggleButtonColor = (uncheckedColor: string) =>
+    isChecked ? 'white' : uncheckedColor;
```
kristojorg: One thing I wonder is whether the `Button` should have this toggle logic, or whether this should only be on a `ToggleButton`... We could put this logic to change the styles directly in the `ToggleButton` component as passing a `bg` or `color` prop will override the ones from the theme. Or we could add a `theme.components.ToggleButton` value with the styles for the toggle button. Not totally sure yet, but I don't think the "checked" logic needs to live in the normal button

kristojorg (changes requested):
A few things I think we can work on next week, but looking really good!

src/ui/Toggle.tsx
siberry: I'm working on this part. Seems a bit tricky with how the `getRadioProps` api works. Seems like we might have to iterate over the children and use `React.cloneElement`?

kristojorg: I think that sounds sensible to me

kristojorg:
I'm going to go ahead and merge this and we can resolve any other issues later

PR #6, opened 2021-06-25
Add baseline NYPL chakra theme

kristojorg:
This PR adds the in-progress NYPL Chakra theme as a baseline from which we can extend. Eventually, we will import this theme from the NYPL Design System, but as it is not available there yet, I have copied it over into the repository. We will want to update values from it to match the designs we have, and can do that in `src/ui/theme`, which extends from the theme defined in `src/ui/nypl-base-theme`.


Toxiapo (approved):

kristojorg:
BTW we can ignore the two deployment failures, this is unrelated and fixed in a separate PR

PR #9, opened 2021-06-25
Add automatic storybook snapshot testing

kristojorg:
This PR adds a config to automatically snapshot test all storybook stories. You can run the tests with `npm run test` and see the results. They will also be run on every PR to ensure we are aware of all snapshot regressions (in stories at least) when developing.


kristojorg:
Ok I have disabled the snapshot testing for now, but this PR still fixes the storybook build and github workflow for testing

src/ui/ToggleButton.tsx
```diff
@@ -5,7 +5,7 @@ import { Icon, IconNames } from '@nypl/design-system-react-components';
 import Button from './Button';
 
 export interface ToggleButtonProps
-  extends React.ComponentPropsWithRef<typeof ChakraBox> {
+  extends React.ComponentPropsWithoutRef<typeof ChakraBox> {
```
hokei, line 8: was this a thing to make it build, or do toggle buttons explicitly disallow refs?

tests/__snapshots__/storyshots.test.ts.snap
```diff
@@ -0,0 +1,2006 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
hokei: should this snapshot file still be here?

.storybook/preview.jsx
```diff
@@ -10,7 +10,7 @@ export const parameters = {
   actions: { argTypesRegex: '^on.*' },
 };
 
-const withChakra = (StoryFn: Function, context: StoryContext) => {
+const withChakra = (StoryFn, _context) => {
```
hokei, line 13: Does it hurt to keep the typings in here?

kristojorg: Yea because I had to switch it to js for storybook to build. They have some kind of bug in their build script with typescript files right now I guess

tests/__snapshots__/storyshots.test.ts.snap
kristojorg: We could remove but it's not hurting anything really

src/ui/ToggleButton.tsx
kristojorg: It doesn't currently accept a ref because we don't use `React.forwardRef`, but if we change it to use that then we can accept a ref here

PR #10, opened 2021-06-29
Style set up for settings button with chakraui

Toxiapo:
This is a WIP update on setting up the styling for the buttons that are going to implemented to the Settings card component.


src/ui/SettingsButton.tsx
```diff
@@ -45,15 +45,23 @@ export default function SettingsCard({ navigator }: { navigator: Navigator }) {
           <Button onClick={open}>Settings</Button>
         </PopoverTrigger>
         <PopoverContent>
-          <PopoverHeader fontWeight="semibold">Settings</PopoverHeader>
+          <PopoverHeader fontWeight="semibold" variant="setting">
+            Settings
+          </PopoverHeader>
```
kristojorg: So in this case I probably wouldn't use the variant functionality of chakra. The way I think about it is that the variant stuff is for when you're going to be using a set of styles with a component over and over again, so I think in this case we should just pass the styles to customize the `PopoverHeader` directly to the component

src/ui/SettingsButton.tsx
```diff
@@ -45,15 +45,23 @@ export default function SettingsCard({ navigator }: { navigator: Navigator }) {
...
           <PopoverCloseButton />
           <PopoverBody>
             <ToggleGroup defaultValue="publisher">
-              <ToggleButton value="publisher">Publisher</ToggleButton>
-              <ToggleButton value="serif">Serif</ToggleButton>
-              <ToggleButton value="sans-serif">Sans-Serif</ToggleButton>
-              <ToggleButton value="dyslexia-friendly">
+              <ToggleButton value="publisher" variant="setting">
+                Publisher
+              </ToggleButton>
```
kristojorg: I think we are only going to have the toggle button in the settings card, so I would actually pass `variant="toggle"` into the `<Button>` you use in there, instead of passing it every time you render a `ToggleButton`.

src/ui/theme/components/button.ts
```diff
@@ -8,6 +8,7 @@ const ButtonStyle = {
   // styles for different visual variants ("outline", "solid")
   variants: {
     solid: variantSolid,
+    setting: variantSetting,
```
kristojorg: I would rename to toggle variant actually. Now that I think about it more, that seems to make more sense

src/ui/theme/components/button.ts
```diff
@@ -40,38 +57,82 @@ function variantSolid(props: any) {
     case 'sepia':
       return {
         ...common,
+        ...commonSettingButtonStyle,
         bg: 'rgb(246, 236, 217)',
         color: 'rgb(17, 17, 17)',
+        _focus: {
+          bg: 'rgb(246, 236, 217)',
+          color: 'rgb(17, 17, 17)',
```
kristojorg: These magic color values should all be added to the theme under `colors.ui` with some names

src/ui/theme/components/button.ts
```diff
@@ -17,6 +18,22 @@ const ButtonStyle = {
...
 
+const commonSettingButtonStyle = {
+  color: 'black',
+  px: '8',
+  border: '1px solid',
+  borderColor: 'gray.100',
+  transition: 'none',
+  maxWidth: '100%',
+  _focus: {
+    'z-index': 1,
```
kristojorg: Is this necessary?

src/ui/theme/components/button.ts
kristojorg: If the values don't make sense in the theme, then they should be declared at the top of the file so they can be reused and edited easily: `const activeGrey = "rgb(...)";`

kristojorg (changes requested):
The code looks great! Nice work : ). I still apparently have something I need to fix with storybook (it hates me) so I haven't seen the preview yet, but I added a few code comments in the mean time.

kristojorg:
Ok after hours of struggling I pushed a dumb commit that fixed the storybook problems 😓 . And now that I can see the preview, I have a couple thoughts:

1. When hovering, there should always be the little hand pointer (`cursor: pointer` in css).
2. Font color should be black (or otherwise match existing UI) when it is not disabled. 
3. The checkmark is a little big, but I think we should just leave it for now. 

So I think the remaining things to be done are:

- [x] Address the code comments above
- [x] Fix hover cursor
- [x] Fix font color
- [ ] Write a couple tests for the toggle button to make sure it shows the checkmark and is accessible and functional

Then we can get this merged : )

kristojorg:
Could you do me a favor and make a storybook story for the `SettingsButton` that shows it with the popover and everything set up?

Edit: Scratch that, this PR is just for the button. We will do that in the PR for the full settings card

src/ui/theme/components/button.ts
```diff
@@ -17,6 +18,32 @@ const ButtonStyle = {
   },
 };
 
+const sepia = 'rgb(246, 236, 217)';
+const lightGreen = '#468028';
+const lightGrey = '#DADADA';
+const defaultBlack = '#383535';
```
kristojorg, line 24: These colors should be pulled from the theme wherever possible. I think one thing that would be helpful would be to make a storybook story that renders all the colors in the theme so you can quickly reference them. I'll go ahead and make a PR that does that.

src/ui/theme/components/button.ts
```diff
@@ -17,6 +18,32 @@ const ButtonStyle = {
...
+const activeBlack = '#111';
+
+const commonToggleButtonStyle = {
+  color: activeBlack,
+  px: '8',
+  border: '0.0625rem solid',
```
kristojorg, line 30: Curious why the `rem` size on the border instead of 1px?

src/ui/theme/components/button.ts
```diff
@@ -17,6 +18,32 @@ const ButtonStyle = {
...
 
+const sepia = 'rgb(246, 236, 217)';
+const lightGreen = '#468028';
+const lightGrey = '#DADADA';
+const defaultBlack = '#383535';
+const activeBlack = '#111';
+
+const commonToggleButtonStyle = {
+  color: activeBlack,
+  px: '8',
```
kristojorg, line 29: All raw numbers in chakra styles can be numbers instead of a string

tests/ToggleButton.test.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { render, screen } from '@testing-library/react';
+import ToggleButton from '../src/ui/ToggleButton';
+
+describe('ToggleButton', () => {
+  test('Button does not show checkmark if isChecked=false', () => {
+    render(<ToggleButton>hello</ToggleButton>);
```
kristojorg, line 7: As an introduction to the testing, I would suggest reading through the react-testing-library docs, which give a pretty good intro to how to think about this stuff.

Read through the intro and guiding principles here: https://testing-library.com/docs/

And then read through the intro and examples in the frameworks/react section. A lot of info, but good to start getting familiar with this stuff!

hokei:
This might be more of a question for @kristojorg - https://web-reader-storybook-git-feature-settings-card-nypl.vercel.app/?path=/story/togglebutton--default&globals=measureEnabled:false why is `isChecked` a prop, rather than a thing handled natively?

kristojorg (approved):
I would still like to pull the colors out into the theme, but that will also be a follow up PR once I render the colors in storybook

PR #12, opened 2021-07-01
add tests to Button and ToggleGroup component

Toxiapo:


tests/ToggleGroup.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return render(
+    <ToggleGroup defaultValue="sedona">
+      <ToggleButton value="sedona">Sedona</ToggleButton>
+      <ToggleButton value="santa_fe">Santa Fe</ToggleButton>
+      <ToggleButton value="las_cruces">Las Cruces</ToggleButton>
+    </ToggleGroup>
+  );
+};
+
+describe('render toggle group component', () => {
```
kristojorg: We don't need this describe wrapper. They're just for grouping tests together, and each file acts as a group as well, so having a single describe isn't necessary. You can just write the tests directly without wrapping them in `describe`

tests/ToggleGroup.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  test('toggle group should contain all radio buttons', () => {
+    const { queryAllByRole } = renderComponent();
+    expect(queryAllByRole('radio')).toHaveLength(3);
+  });
+
+  test('triggering a click should alter button state', () => {
+    const { getByLabelText } = renderComponent();
+
+    expect(getByLabelText('Sedona')).toBeChecked();
```
kristojorg: `getByLabelText` is fine here, but a slightly stronger test would be `getByRole("radio", {name: "Sedona"})` since it enforces both the `role` and the `aria-label` attributes. But it's all good, no need to change just a thought for the future

tests/Button.test.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import { render, screen } from '@testing-library/react';
+import Button from '../src/ui/Button';
+
+describe('render button component', () => {
```
kristojorg: We don't need this wrapping `describe`

tests/ToggleGroup.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const { getByRole } = renderComponent();
+    expect(getByRole('radiogroup')).toBeInTheDocument();
+  });
+
+  test('toggle group should contain all radio buttons', () => {
+    const { queryAllByRole } = renderComponent();
+    expect(queryAllByRole('radio')).toHaveLength(3);
+  });
+
+  test('triggering a click should alter button state', () => {
```
kristojorg: The way we will use this component is as a "controlled" component, meaning we will pass it a `value` prop and it should respect that value prop and simply call `onChange` when a click is registered. So this functionality of managing its own state is actually something we don't want. Instead, we should test that:

1. When we pass a `value` to the group, the right button is checked
2. When a button is clicked, `onChange` fires with the right value (you already test this below)

tests/ToggleGroup.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <ToggleGroup defaultValue="sedona" onChange={onChangeHandler}>
+        <ToggleButton value="sedona">Sedona</ToggleButton>
+        <ToggleButton value="santa_fe">Santa Fe</ToggleButton>
+        <ToggleButton value="las_cruces">Las Cruces</ToggleButton>
+      </ToggleGroup>
+    );
+
+    fireEvent.click(getByLabelText('Santa Fe'));
+
+    expect(onChangeHandler).toHaveBeenCalledTimes(1);
```
kristojorg: You can also test that it was called with `santa_fe` as the first argument using `toHaveBeenCalledWith`

kristojorg (changes requested):
This looks good, just a couple small changes!

tests/ToggleGroup.test.tsx
Toxiapo: I will remove it and also the test for the Button, thanks!

tests/ToggleGroup.test.tsx
Toxiapo: Updated, thanks!

kristojorg (approved):

PR #15, opened 2021-07-07
add aria-lable to radio group

Toxiapo:
This PR adds the `label` prop to the `radiogroup` component. `label` should be a required prop.


kristojorg (approved):
Looks great to me, LGTM!

PR #16, opened 2021-07-08
New D2Reader API

kristojorg:
This PR installs `@d-i-t-a/reader@2.0.0-alpha.2` and implements their new API. It also resolves the problem we were having sometimes with `this` having the wrong value in methods by binding `this` to the class in the constructor.

This is not an ideal solution to the problem of `this`, but we will ultimately resolve the remaining problems by switching to a redux-style state management instead of using these classes the way we are.

### DIFF
--- a/src/HtmlNavigator/HtmlNavigatorContent.tsx
+++ b/src/HtmlNavigator/HtmlNavigatorContent.tsx
@@ -3,125 +3,15 @@ import * as React from 'react';
 const EpubContent = () => {
   return (
     <div style={{ height: '100%', overflow: 'hidden' }}>
-      <div id="root">
-        <div
-          id="D2Reader-Container"
-          style={{ width: '100%', height: '100%', position: 'relative' }}
+      <div id="D2Reader-Container">
+        <main
+          style={{ height: 'calc(100vh - 150px)' }}
+          tabIndex={-1}
+          id="iframe-wrapper"
         >
-          <main
-            style={{ overflow: 'hidden' }}
-            tabIndex={-1}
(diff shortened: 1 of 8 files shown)


src/HtmlNavigator/HtmlNavigatorContent.tsx
```diff
@@ -3,125 +3,15 @@ import * as React from 'react';
...
-                <i className="material-icons" style={{ color: 'gray' }}>
-                  delete
-                </i>
-              </button>
-            </div>
-          </span>
-        </div>
+          <div id="reader-loading" className="loading"></div>
+          <div id="reader-error" className="error"></div>
+        </main>
```
kristojorg, line 14: The new D2Reader release also included some fixes that allowed us to clean up our HTML for the HTMLNavigator

Toxiapo (approved):

src/HtmlNavigator/HtmlNavigatorContent.tsx
hokei: does this mean the "highlight-toolbox" classes, etc, are automatically injected?  Or were these just things we had to include because of D2 but we didn't actually need to use them?

hokei:
why is it that github is showing lint errors but the lint check is passing?

src/HtmlNavigator/HtmlNavigatorContent.tsx
kristojorg: Things we had included before but didn't need

kristojorg:
I think those are just lint warnings

PR #17, opened 2021-07-08
Sfr 1169 settings card

Toxiapo:
This PR adds the final styles and the A-/A+ options to the settings card popup.


kristojorg:
The code here looks good to me, but a couple questions:

- Can we add some shadow to the settings popover so that it is more distinct from the background?
- ~The fonts in the buttons appear to be incorrect. I think we should be using system font by default, can you double check we are indeed doing that in the theme, and maybe try to figure out how this is being set to font "Times" ?~
- The settings card should close when the user clicks anywhere outside of it

Edit: I figured out we weren't actually rendering a `<ChakraProvider>` in the example app anywhere, so I added that to the `index.tsx` file and that fixed the font issue. It also caused a couple other style things that I cleaned up for us.

kristojorg:
PS, the example deployment is not currently working. I have fixed that in my PR separately. It'll work again once we merge that one

kristojorg (changes requested):
Just need some UI changes noted above

Toxiapo:
> * Can we add some shadow to the settings popover so that it is more distinct from the background?

Sure! I will add some background color to the popover.

Toxiapo:
> * The settings card should close when the user clicks anywhere outside of it

I can actually remove the [open/close](https://github.com/NYPL-Simplified/web-reader/blob/bbbfba5fbb304c629716004723a1f3cf5c195eb8/src/ui/SettingsButton.tsx#L14) state and this should work. Unless there's a reason that we need this component to be controlled?

kristojorg:
I don't remember why I had it controlled in the first place... I vaguely remember there being a reason, but if removing that makes it work the way we want then let's do it

kristojorg:
> Sure! I will add some background color to the popover.

I was thinking more like a box-shadow so that it looks like it's hovering over the content a bit

src/ui/SettingsButton.tsx
```diff
@@ -43,18 +43,19 @@ export default function SettingsCard({ navigator }: { navigator: Navigator }) {
...
         <PopoverTrigger>
           <Button onClick={open}>Settings</Button>
         </PopoverTrigger>
         <PopoverContent
           borderColor="gray.100"
           width="fit-content"
-          _active={{ boxShadow: 'none' }}
+          _active={{ boxShadow: 'lg' }}
+          _focus={{ boxShadow: 'lg' }}
+          boxShadow="lg"
```
kristojorg: Do you need to have this on `_active` and `_focus`? If you just have it without the pseudoselectors does it not display correctly?

kristojorg:
There seems to be an issue with the focus or tabindex in the page... for some reason I cannot tab onto the settings button, and then when it is open, I can't tab between the options inside. I don't know that this happened because of this PR, but either way we need to fix it as it is an accessibility issue

src/ui/SettingsButton.tsx
Toxiapo: I believe we do need to explicitly define them if we want to be consistent with the boxShadow style. The default one was a blurred blue border.

kristojorg:
Ok I did a little looking and it looks like you are overriding the `_focus` pseudoselector in `theme/button.ts`. This probably does match the old webpub viewer, but it's not accessiblity compliant and thus we should undo that and leave the default chakra focus ring on all buttons

Toxiapo:
> over



> There seems to be an issue with the focus or tabindex in the page... for some reason I cannot tab onto the settings button, and then when it is open, I can't tab between the options inside. I don't know that this happened because of this PR, but either way we need to fix it as it is an accessibility issue

Yea I see what you mean, it's giving me the same problem.

kristojorg:
Update: you may even need to proactively add the focus state to the button, because underneath the custom buttons we really want to be tabbing into the `<input>` I believe. So you tab through the inputs, but the button displays the focus state...

src/ui/SettingsButton.tsx
kristojorg: I think we shouldn't ever remove the focus ring, if the element is focused it should have the focus ring. There are times when we can remove it if the user isn't actively tabbing through the site, but we shouldn't get that fancy yet. 

I think we should set the initial focus when opening the popover to the first choice the user has. You can see an example of that here:

https://chakra-ui.com/docs/overlay/popover#focus-an-element-when-popover-opens

Toxiapo:
> There seems to be an issue with the focus or tabindex in the page... for some reason I cannot tab onto the settings button, and then when it is open, I can't tab between the options inside. I don't know that this happened because of this PR, but either way we need to fix it as it is an accessibility issue

After doing some debugging with @hokei, this could potentially be an issue coming from chakraui. Removing the [scheduleMicrotask](https://github.com/chakra-ui/chakra-ui/blob/db46840efea41d66c4d7c8f59af7d944dfb0cbd9/packages/radio/src/use-radio.ts#L218) wrapper function fixes it. 

We should probably still need to do more investigation on this.

Edit: Follow up with chakra-ui/chakra-ui#4259

kristojorg:
Ok, if @hokei approves the code/preview I would say we merge this and make a new ticket to create a PR to chakra with that fix and ensure this issue is resolved in our code once that's done

hokei (approved):
Just those little lint issues, and  let's create a follow-up ticket for the focus bug.

Toxiapo:
> Ok, if @hokei approves the code/preview I would say we merge this and make a new ticket to create a PR to chakra with that fix and ensure this issue is resolved in our code once that's done

We are ready to close this for now.

PR #18, opened 2021-07-09
A more Reacty approach

kristojorg:
This PR makes a significant change to the architecture of the project, moving us from using `HtmlNavigator` and `PdfNavigator`s internally, to using hooks: `usePdfReader` and `useHtmlReader`. This approach is going to be much simpler long-term because it fits more naturally into the React ecosystem.

In making this change, I also got the color mode and pagination buttons to fully work : ). They'll even remember your previous setting when you refresh the page.

This also provides us a path forward to creating an entirely react-based html reader, where we ultimately don't depend on D2Reader at all, but just perform the necessary logic ourselves in our own hook, which I think will me _much_ simpler in the long-run. 

Note: To view it, use the "Regular (streamed) example" in the preview app
Note 2: there appears to be a bug with the color of the reader in safari (only in a production build). We need to make a new ticket for this.


README.md
```diff
@@ -95,26 +107,26 @@ Finally, to use in a vanilla Javascript app:
...
 
 **Notes:**
-- There is one `Navigator` per _media-type_ (PDF, HTML, Image, etc), not per _format_. As in, ePub and Mobi books are different formats that use the same media type (HTML). Audiobooks and PDF collections use different media types. We currently only have plans for HTML and PDF, but other Navigators are welcome and should fit right in.
+- There is one `use_X_Reader` per _media-type_ (PDF, HTML, Image, etc), not per _format_. As in, ePub and Mobi books are different formats that use the same media type (HTML). Audiobooks and PDF collections use different media types. We currently only have plans for HTML and PDF, but other hooks are welcome and should fit right in.
 - We always start from a Webpub Manifest. This means other formats (like ePub) need to be processed before they get to us. This can be done with a Readium Streamer, or some other way. 
   - For example, DRB is pre-generating PDF manifests from web-scraped content.
```
hokei, line 115: you can link to that repo here, I think it's also open source

Toxiapo (commented):
Is the Color mode fix somewhat connected to what I am about to do for the color mode functionality ticket?

Toxiapo (approved):

kristojorg:
@Toxiapo 
> Is the Color mode fix somewhat connected to what I am about to do for the color mode functionality ticket?

No this was a separate issue where the mode wasn't being applied to the iframe at all. The color mode for the ui elements is separate

hokei (approved):
wow useReducer is much nicer than redux haha

PR #20, opened 2021-07-09
update default active state style

Toxiapo:
This PR fixes incorrect button style in active state.

### DIFF
--- a/src/ui/theme/components/button.ts
+++ b/src/ui/theme/components/button.ts
@@ -42,6 +42,15 @@ const commonToggleButtonStyle = {
   _active: {
     bg: 'ui.white',
   },
+  _hover: {
+    bg: 'ui.white',
+    _disabled: {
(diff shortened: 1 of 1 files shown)


kristojorg:
Sepia disabled and dark disabled appear to share similar issues, but this looks good on the light one

Toxiapo:
> Sepia disabled and dark disabled appear to share similar issues, but this looks good on the light one

Updated.

I had to redefine the pseudoselectors styles for the other two buttons. Is this what we should be doing?

hokei (approved):

kristojorg:
> I had to redefine the pseudoselectors styles for the other two buttons. Is this what we should be doing?

Yea I don't think there is any way around this really. I'm pretty sure we would have to do this with normal css too. The button has some defaults but they don't apply to every variant

PR #21, opened 2021-07-13
Implement prev/next buttons

Toxiapo:
This PR adds the Next and Previous buttons to the page.

Couple of things that we still need to take care of:

- On `Scroll` mode, the buttons should be hidden. Do we add an additional prop to the button or this could be handled in the parent component?
- Still need to fix the color/background when we introduce color mode.

### DIFF
--- a/src/ui/PageButton.tsx
+++ b/src/ui/PageButton.tsx
@@ -16,9 +16,11 @@ function PageButton(props: React.PropsWithoutRef<PageButtonProps>) {
         alignItems="center"
         border="1px solid"
         borderColor="ui.gray.light-cool"
-        bg="ui.white"
-        mx={7}
-        size="50px"
+        color="ui.black"
+        bg="inherit"
+        size={{ sm: '40px', md: '50px' }}
+        mr={8}
+        ml={6}
         cursor="unset"
         _hover={{ background: 'ui.black', color: 'ui.white' }}
         onClick={onClick}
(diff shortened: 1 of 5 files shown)


kristojorg:
I think for now let's assume the parent component just won't render the prev/next at all in scroll mode. 

We can also come back to color mode all at once in that color mode ticket. I'd like to see if we can use the built in chakra support for color modes while adding the third sepia option

src/ui/PageButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+        alignItems="center"
+        border="1px solid"
+        borderColor="ui.gray.light-cool"
+        color="ui.black"
+        bg="inherit"
+        size={{ sm: '40px', md: '50px' }}
+        mr={8}
+        ml={6}
+        cursor="unset"
+        _hover={{ background: 'ui.black', color: 'ui.white' }}
```
hokei: I think the color on the reader is like, #4d4d4d.  Do we have a dark grey preset?

hokei (changes requested):
Just one nit, and the lint warnings, which should be easy enough to remove and will make things cleaner.

src/ui/PageButton.tsx
Toxiapo: Yes, we do! Thank you.

hokei (approved):

PR #23, opened 2021-07-16
Sfr 1201 header bar

Toxiapo:
This PR adds the header bar design to the page.

**NOTE**
- All the icons are temporary until we finalize what icons to use.
- Still missing the TOC button

### DIFF
--- a/src/ui/Header.tsx
+++ b/src/ui/Header.tsx
@@ -21,7 +21,14 @@ export default function Header(props: HeaderProps) {
   const { logo, readerState, navigator } = props;
 
   return (
-    <Flex alignContent="space-between" alignItems="center" py={1} px={8}>
+    <Flex
+      alignContent="space-between"
+      alignItems="center"
+      py={2}
+      px={8}
+      border="1px solid"
+      borderColor="gray.100"
+    >
       <Link
         href="/"
         aria-label="Return to Digital Research Books"
--- a/src/ui/Header.tsx
+++ b/src/ui/Header.tsx
@@ -25,15 +25,15 @@ export default function Header(props: HeaderProps) {
       alignItems="center"
       py={2}
       px={8}
-      border="1px solid"
+      borderBottom="1px solid"
       borderColor="gray.100"
     >
       {headerLeft ? (
         headerLeft
       ) : (
         <Link
-          href="/"
-          aria-label="Return to Digital Research Books"
+          href="https://www.nypl.org"
+          aria-label="Return to NYPL"
           tabIndex={0}
           fontSize={0}
           py={1}
@@ -46,7 +46,7 @@ export default function Header(props: HeaderProps) {
           }}
         >
           <Icon decorative name={IconNames.headset} modifiers={['small']} />
-          <Text variant="headerNav">Return to Digital Research Books</Text>
+          <Text variant="headerNav">Return to NYPL</Text>
         </Link>
       )}
       <HStack ml="auto" spacing={1}>
--- a/src/ui/Header.tsx
+++ b/src/ui/Header.tsx
@@ -1,0 +1,55 @@
+import React from 'react';
+import SettingsCard from './SettingsButton';
+import {
+  Flex,
+  Link,
+  HStack,
+  Button as ChakraButton,
+  Text,
+} from '@chakra-ui/react';
+import { Icon, IconNames } from '@nypl/design-system-react-components';
+import { WebpubManifest, ReaderState, Navigator } from '../types';
+
+export type HeaderProps = {
+  logo?: any;
+  manifest: WebpubManifest; // for TOC?
+  readerState: ReaderState;
+  navigator: Navigator;
+};
+
+export default function Header(props: HeaderProps) {
+  const { logo, readerState, navigator } = props;
+
+  return (
+    <Flex alignContent="space-between" alignItems="center" py={1} px={8}>
+      <Link
+        href="/"
+        aria-label="Return to Digital Research Books"
+        tabIndex={0}
+        fontSize={0}
+        py={1}
+        textTransform="uppercase"
+        color="gray.700"
+        d="flex"
+        alignItems="center"
+        _hover={{
+          textDecoration: 'none',
+        }}
+      >
+        {logo ? (
+          logo
+        ) : (
+          <Icon decorative name={IconNames.headset} modifiers={['small']} />
+        )}
+        <Text variant="headerNav">Return to Digital Research Books</Text>
+      </Link>
+      <HStack ml="auto" spacing={1}>
+        <SettingsCard navigator={navigator} state={readerState} />
+        <ChakraButton variant="headerNav">
+          <Icon decorative name={IconNames.search} modifiers={['small']} />
+          <Text variant="headerNav">Toggle Fullscreen</Text>
+        </ChakraButton>
+      </HStack>
+    </Flex>
+  );
+}
--- a/src/ui/Header.tsx
+++ b/src/ui/Header.tsx
@@ -8,17 +8,16 @@ import {
   Text,
 } from '@chakra-ui/react';
 import { Icon, IconNames } from '@nypl/design-system-react-components';
-import { WebpubManifest, ReaderState, Navigator } from '../types';
+import { ReaderState, Navigator } from '../types';
 
 export type HeaderProps = {
-  logo?: any;
-  manifest: WebpubManifest; // for TOC?
+  headerLeft?: any; // Top-left header section
   readerState: ReaderState;
   navigator: Navigator;
 };
 
 export default function Header(props: HeaderProps) {
-  const { logo, readerState, navigator } = props;
+  const { headerLeft, readerState, navigator } = props;
 
   return (
     <Flex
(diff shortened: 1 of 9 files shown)


src/ui/Header.tsx
```diff
@@ -0,0 +1,62 @@
...
+  HStack,
+  Button as ChakraButton,
+  Text,
+} from '@chakra-ui/react';
+import { Icon, IconNames } from '@nypl/design-system-react-components';
+import { WebpubManifest, ReaderState, Navigator } from '../types';
+
+export type HeaderProps = {
+  logo?: any;
+  manifest: WebpubManifest; // for TOC?
```
hokei: I'd say get rid of this for now.  It's currently not being used at all (L21) and we can add it back in when we know what we need it for

src/ui/Header.tsx
```diff
@@ -0,0 +1,62 @@
...
+      alignContent="space-between"
+      alignItems="center"
+      py={2}
+      px={8}
+      border="1px solid"
+      borderColor="gray.100"
+    >
+      <Link
+        href="/"
+        aria-label="Return to Digital Research Books"
```
hokei: We don't want to hardcode this (and also L51)

src/ui/Header.tsx
```diff
@@ -0,0 +1,62 @@
...
+  Link,
+  HStack,
+  Button as ChakraButton,
+  Text,
+} from '@chakra-ui/react';
+import { Icon, IconNames } from '@nypl/design-system-react-components';
+import { WebpubManifest, ReaderState, Navigator } from '../types';
+
+export type HeaderProps = {
+  logo?: any;
```
hokei: I wouldn't make this separate from the "Return to Digital Research Books" text - just let anyone pass whatever they want.  Not sure what we'd call this variable but if we put a comment saying it's the thing on the top left I'm sure people will understand.

src/ui/Header.tsx
Toxiapo: That makes sense. Updated.

src/ui/SettingsButton.tsx
```diff
@@ -26,14 +28,18 @@ export default function SettingsCard({
...
+        placement="bottom"
         closeOnBlur={false}
         isOpen={isOpen}
         onClose={close}
         autoFocus={false}
       >
         <PopoverTrigger>
-          <Button onClick={open}>Settings</Button>
+          <ChakraButton onClick={open} variant="headerNav">
+            {' '}
```
hokei: nit: are these necessary?

tests/Header.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+import Header from '../src/ui/Header';
+import { Navigator, ReaderState } from '../src/types';
+
+test('render header bar', () => {
+  const readerState = {} as ReaderState;
+  const navigator = {} as Navigator;
+
+  render(<Header readerState={readerState} navigator={navigator} />);
+
+  // Should also test the href once everthings been implemented in the furture
```
hokei: Additional tests:  
 Mount and pass a custom thing headertext and icon and make sure it shows up.

tests/Header.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+  // By default it should show the NYPL logo and a link
+  expect(
+    screen.getByRole('link', { name: 'Return to Digital Research Books' })
```
hokei: Let's make this default "Return to NYPL" instead of "Digital Research Books" and make the href `www.nypl.org`

tests/Header.test.tsx
Toxiapo: I added a test using `data-testid` for the icon b/c I can't target the Icon. Would love to know if there are other ways to go for this.

src/ui/manager.tsx
```diff
@@ -26,21 +21,9 @@ const ManagerUI: React.FC<ReaderReturn> = ({
...
-          }}
-        >
-          <h1>{manifest?.metadata.title}</h1>
-          <div>
-            {navigator && state && (
-              <SettingsCard navigator={navigator} state={state} />
-            )}
-          </div>
-        </nav>
+        {navigator && state && (
```
hokei, line 24: What does it mean when either navigator or state are not present here?  do we need to throw an error?

src/ui/theme/components/button.ts
```diff
@@ -9,6 +9,7 @@ const ButtonStyle = {
   variants: {
     solid: variantSolid,
     toggle: variantToggle,
+    headerNav: variantHeaderNav,
```
hokei, line 12: why is this a new variant rather than just a solid button?

src/ui/theme/components/button.ts
```diff
@@ -151,4 +152,15 @@ function variantToggle(props: any) {
   };
 }
 
+function variantHeaderNav() {
+  return {
+    py: 1,
+    border: 'none',
+    background: 'transparent',
```
hokei: what is the difference between this and `bg` used in other variants?  Why use 'transparent' rather than white/sepia/dark ?

hokei:
Screen Shot 2021-07-19 at 2 34 54 PM

Screenshot below from digital-research-books-reader
Screen Shot 2021-07-19 at 2 36 00 PM


Might also be worth it to add a `colorScheme` control to `header` stories, just like in `toggleButton`

tests/Header.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+  expect(
+    screen.getByRole('button', { name: 'Toggle Fullscreen' })
+  ).toBeInTheDocument();
+});
+
+test('render custom left header', () => {
+  const readerState = {} as ReaderState;
+  const navigator = {} as Navigator;
+  const IconComponent = () => (
+    <span data-testid="custom-icon">
```
hokei, line 33: It's totally find to use testId here because it's scoped just to the test.  It's only really a smell when you're finding yourself putting testIds on stuff outside of the test, such as the "toggle fullscreen" button

hokei:
Clicking outside the settings box doesn't cause the settings box to close

I'm also having trouble tabbing from outside the settings box into the settings box. This might have to do with that focus bug. 

Since this is a Header visuals ticket, I wonder if it's worth it to create a new ticket that is something like "integrate the header into webreader", where we deal with these issues.  Let me know what you think - I don't want to balloon this ticket, but if it's a small fix we can fix it here.

hokei:
I'm also having trouble tabbing from outside the settings box into the settings box

hokei (changes requested):
Some more things to change - some of these might be better to pull out into new tickets, let me know what you think.

src/ui/theme/components/text.ts
```diff
@@ -0,0 +1,24 @@
...
+  sizes: {},
+  // styles for different visual variants ("outline", "solid")
+  variants: {
+    headerNav: variantHeaderNav,
+  },
+  // default values for `size`, `variant`, `colorScheme`
+  defaultProps: {},
+};
+
+function variantHeaderNav() {
```
hokei, line 14: why have this defined here rather than in styled components?  Why use this only for "settings" and "toggle fullscreen" but not the "Return to NYPL" which visually looks the same?  Is this a Chakra best practice?

src/ui/theme/components/button.ts
Toxiapo: Whoops, `bg` == `background` in chakraui, I will update it so it matches the rest of the code. 

We have a separate style for the toggle buttons and the rest of the elements.`white/sepia/dark` are the hardcoded colors specifically for the three buttons on the Settings Card, the HeaderNav background is dynamic and will change based on the color mode setting. (Not yet implemented)

src/ui/theme/components/button.ts
Toxiapo: solid button contains styles that are specifically for the toggle buttons.

src/ui/manager.tsx
Toxiapo: It means the App is still loading the content, [initially null](https://github.com/NYPL-Simplified/web-reader/blob/48f0854796577e8514a6b0639504d273c8a0e1ce/src/useWebReader.tsx#L83)

src/ui/theme/components/text.ts
Toxiapo: This is also used for **Return to NYPL**, it's a `<Text />` variant. I created the variant b/c it's being used in multiple places.

Toxiapo (commented):
> Clicking outside the settings box doesn't cause the settings box to close
> 
> I'm also having trouble tabbing from outside the settings box into the settings box. This might have to do with that focus bug.
> 
> Since this is a Header visuals ticket, I wonder if it's worth it to create a new ticket that is something like "integrate the header into webreader", where we deal with these issues. Let me know what you think - I don't want to balloon this ticket, but if it's a small fix we can fix it here.

The merge from the main should solve the issues, but let me know if the bug is still appearing.

hokei:
> solid button contains styles that are specifically for the toggle buttons.

Will the header nav buttons and the toggle buttons always match?  If yes, we should make them the same thing. But if not, we can keep them separate.

hokei:
> The merge from the main should solve the issues, but let me know if the bug is still appearing.

Seems like clicking outside the settings box is still not causing the box to close.  However, the focus thing is fixed.

Toxiapo:
> > The merge from the main should solve the issues, but let me know if the bug is still appearing.
> 
> Seems like clicking outside the settings box is still not causing the box to close. However, the focus thing is fixed.

Hmm, Chrome works, but not FF, really weird...

Toxiapo:
> Might also be worth it to add a `colorScheme` control to `header` stories, just like in `toggleButton`

I will create a ticket to add colorScheme.

hokei (approved):
Add a new ticket for the FF click issue as well, and then let's get this in!

src/ui/theme/components/button.ts
kristojorg: I don't think this needed to be a new variant necessarily, just a separate component that overrides some of the basic button styles. The way I think about this is: If you need a common component with some specific styling, is it going to be re-used all over the app, or is it just for a well defined area? If it is going to be re-used all over then make it a variant or new component in the theme... otherwise, just make a component called `HeaderButton` that renders the `<Button>` underneath and overrides whatever styles you need. I think theme variants/components should really just be for the most common stuff

Toxiapo: Ok, that makes sense. I will include this in my next update. Thanks!

PR #25, opened 2021-07-27
SFR-1222 pdf renderer

hokei:
Adds PDF renderer to web-reader

Currently just has page-turning functionality.  TOC, scrolling/paginated, etc will be added in follow up tickets.  

I feel like this needs some tests, but I'm not sure how to go about this.


hokei:
@kristojorg Just wanted to get your eyes on this, and let me know if what I've done in /PdfReader/index.tsx seems reasonable?

example/index.tsx
```diff
@@ -34,7 +34,7 @@ const AxisNowEpub = () => {
 const SWTest = () => {
   const [value, setValue] = React.useState<null | string>(null);
   React.useEffect(() => {
-    fetch('http://localhost:1234/samples/muse13454.json').then((res) => {
```
kristojorg, line 37: Did you intend to edit the SWTest example for now? It's just a little playground for me to work on the SW basically so wouldn't be that useful to you right now I don't think

src/PdfReader/index.tsx
```diff
@@ -30,26 +33,48 @@ function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
...
   const { webpubManifestUrl, manifest } = args ?? {};
-  const [state] = React.useReducer(pdfReducer, {
+
+  const [state, dispatch] = React.useReducer(pdfReducer, {
     type: 'PDF',
     colorMode: 'day',
     isScrolling: false,
     fontSize: 16,
     fontFamily: 'sans-serif',
+    resource: '',
```
kristojorg: My personal preference would probably be to use `null` before any resource has been set as I think it more explicitly communicates what's happening and leaves less room for bugs, especially since `''` is seen as a valid string just like a resource url would be

src/PdfReader/index.tsx
```diff
@@ -30,26 +33,48 @@ function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
...
+  const [numPages, setNumPages] = useState(0);
+  const [isLoading, setLoading] = useState(true);
+
   // initialize the pdf reader
   React.useEffect(() => {
+    console.log('initializing', manifest);
     // bail out if there is not manifest passed in,
     // that indicates that this format is inactive
     if (!manifest) return;
+    const resource: string = manifest.readingOrder![0].href;
```
kristojorg: I wouldn't force unwrap this, but probably throw an error if the reading order has nothing in it, which we can later handle further up the component tree

src/PdfReader/index.tsx
```diff
@@ -30,26 +33,48 @@ function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
...
     colorMode: 'day',
     isScrolling: false,
     fontSize: 16,
     fontFamily: 'sans-serif',
+    resource: '',
+    pageNumber: 1,
   });
 
+  const [numPages, setNumPages] = useState(0);
+  const [isLoading, setLoading] = useState(true);
```
kristojorg: I think you can derive the `isLoading` state based on whether `resource` is `null` in the state. That's what I would do. I also think I would store the `numPages` in the state with everything else. I don't think you should need `useState` at all since we have `useReducer`

src/types.ts
```diff
@@ -35,6 +35,13 @@ export type WebpubManifest = {
   spine: Spine<'text/html'>;
   links: any[];
   resources: any[];
+  readingOrder?: ReadingOrder[];
+};
+
+export type ReadingOrder = {
+  href: string;
+  title: string;
+  type: string;
```
kristojorg: This is good, our WebpubManifest type is out of date actually. I generated a full type from the official specs in epub-to-webpub that I'm going to have Jiayong extract and we can then import it here and it will have full defs for `readingOrder` etc.

src/types.ts
```diff
@@ -11,7 +11,7 @@ export const WebpubMimeType = 'application/webpub';
  * THE FOLLOWING TWO TYPES ARE MADE UP STUBS
  */
 // a webpub pdf collection
-export const WebpubPdfConformsTo = 'stub/webpub+pdf';
+export const WebpubPdfConformsTo = 'application/webpub+pdf';
```
kristojorg: I would leave this as is right now to be clear it's just a stub... eventually it will be replaced with whatever we decide with mike. but not a big deal either

.env.local
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_PROXY_URL=www.google.com
```
kristojorg: .env.local should not be checked in to source control. I'm also not totally sure it'll be picked up right now since I think we need some special setup to use the `.env` files

src/PdfReader/index.tsx
```diff
@@ -74,30 +101,45 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
+      content: <>PDF is loading</>,
       manifest: null,
       navigator: null,
       state: null,
     };
   }
 
+  function onDocumentLoadSuccess({ numPages }: { numPages: number }) {
+    setNumPages(numPages);
+  }
```
kristojorg: I would make a reducer action called `DOCUMENT_LOAD_SUCCESS` and and pass anything we need from here to that to be set inside the `state` variable.

kristojorg (commented):
Ok looks pretty good to me. I made a few small comments, but you're on the right track. One thing we will have to figure out how to deal with eventually is how the state and navigator and stuff will differ between `usePdfReader` and `useHtmlReader` and how to handle that in the UI, but that's not something we need to stress about now. Lookin good to me!

example/index.tsx
hokei: I moved the file, so I just updated everything that referred to the old file.  I don't think this affects anything.

kristojorg: Oh kk

.env.local
hokei: yeah I was being dumb, we're not using next at all here anyway

src/PdfReader/index.tsx
hokei: The presence of "state.data" is a better indicator of load success than anything else - so I removed it.  Let me know if this all makes sense anyway.

Toxiapo (commented):
Looks good! 🥳

src/PdfReader/index.tsx
```diff
@@ -30,41 +51,121 @@ function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
...
     // bail out if there is not manifest passed in,
     // that indicates that this format is inactive
     if (!manifest) return;
-    // here initialize reader however u do
-    // ...
-  }, [manifest]);
+    // throw an error on a badly formed manifest
+    if (!manifest.readingOrder || !manifest.readingOrder.length) {
+      throw new Error('Manifest has no Reading Order');
+    }
```
Toxiapo, line 146: Why are we explicitly checking the ReadingOrder?

hokei: the `ReadingOrder` is the part of the webpub that lists the resources in the order they're supposed to be read.  If there's no reading order, there's something wrong with the manifest.

package.json
```diff
@@ -43,11 +43,11 @@
   "size-limit": [
     {
       "path": "dist/web-reader.cjs.production.min.js",
-      "limit": "600 KB"
+      "limit": "650 KB"
```
hokei, line 46: I've increased the size limit because adding the pdfjs package made it bigger - wondering if it's possible to only import the pdf or the epub packages based on what you need.

kristojorg: Yea we should come back to this at a later date and try to keep this lib small. The R2D2BC package is also quite large so I'll want to fix that up later.

src/PdfReader/index.tsx
```diff
@@ -1,24 +1,45 @@
...
 } from '../types';
-import PDFContent from './PdfContent';
 
 type PdfState = ReaderState & {
   type: 'PDF';
-  // some pdf-specific state here, like the instance of the
-  // package you are using if necessary
+  loadSuccess: boolean;
+  resourceIndex: number;
+  data: { data: Uint8Array } | null;
```
kristojorg: Why are we storing it as `data.data` instead of just `data: Uint8Array | null` ?

src/PdfReader/index.tsx
```diff
@@ -1,24 +1,45 @@
...
+  | { type: 'SET_DATA'; data: { data: Uint8Array } | null }
   | { type: 'SET_COLOR_MODE'; mode: ColorMode }
-  | { type: 'SET_SCROLL'; isScrolling: boolean };
+  | { type: 'SET_SCROLL'; isScrolling: boolean }
+  | { type: 'SET_NUMPAGES'; numPages: number }
+  | { type: 'SET_PAGENUM'; pageNum: number };
 
 function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
   switch (action.type) {
+    case 'SET_RESOURCEINDEX':
```
kristojorg: I think some comments here would be useful to know what these actions are meant to represent

src/PdfReader/index.tsx
```diff
@@ -1,24 +1,45 @@
...
   ReaderState,
+  WebpubManifest,
 } from '../types';
-import PDFContent from './PdfContent';
 
 type PdfState = ReaderState & {
   type: 'PDF';
-  // some pdf-specific state here, like the instance of the
-  // package you are using if necessary
+  loadSuccess: boolean;
```
kristojorg: What does this represent? The success of an individual resource or of the react-pdf package or something else? If this state can be deduced from something else in the state, I would define it that way in the `usePdfReader` instead of in the state. For example does `loadSuccess = !!state.data` ?

src/PdfReader/index.tsx
```diff
@@ -30,41 +51,121 @@ function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
...
+    resourceIndex: number,
+    proxyUrl?: string
+  ) => {
+    // Generate the resource URL using the proxy
+    const resource: string =
+      proxyUrl + encodeURI(manifest.readingOrder![resourceIndex].href);
+
+    return await fetchPdf(resource);
+  };
+
```
kristojorg: If this isn't using anything inside the `usePdfReader` closure I think we should move it out to the file's top level

src/PdfReader/index.tsx
```diff
@@ -30,41 +51,121 @@ function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
...
   // initialize the pdf reader
   React.useEffect(() => {
+    async function setPdfResource(manifest: WebpubManifest, proxyUrl: string) {
+      const data = await loadResource(manifest, 0, proxyUrl);
+      dispatch({ type: 'SET_DATA', data: { data } });
+      dispatch({ type: 'SET_RESOURCEINDEX', index: 0 });
```
kristojorg: This is a matter of preference but I personally would prefer having the actions more describe the things that are happening and not the change to the state. That way we don't end up with a separate action to update every piece of state like this, but instead have `dispatch({type: "RESOURCE_LOADED", data});` and then the reducer sets the data and the index to `0`. The reason I had `SET_SCROLL` and `SET_COLOR_MODE` is because those are actually independent actions that can happen, they just happen to be related to individual pieces of state.

src/PdfReader/index.tsx
```diff
@@ -30,41 +51,121 @@ function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
...
-    console.log('unimplemented');
-  }, []);
+      const data = await loadResource(
+        manifest!,
+        state.resourceIndex + 1,
+        proxyUrl
+      );
+      dispatch({ type: 'SET_DATA', data: { data } });
+      dispatch({ type: 'SET_RESOURCEINDEX', index: state.resourceIndex + 1 });
+    }
```
kristojorg: I would again structure this differently so we don't have one `dispatch` per piece of state we want to change, but one per logical action taking place. I see two possible actions from goForward: `NAVIGATE_NEXT_PAGE` and `NAVIGATE_NEXT_RESOURCE`. The logic to handle those things would then be consolidated into the reducer.

I would then have the `loadResource` function dispatch the actions to set the data to null:
```ts:
async function loadResource(manifest, index, proxyUrl){
  dispatch({type: "LOAD_RESOURCE"});
  const data = await fetch(...)
  dispatch({type: "LOAD_RESOURCE_SUCCESS", data });
}
```

This way the logic about how the state responds to actions is all considated in the reducer and now spread around the various functions

src/PdfReader/index.tsx
```diff
@@ -30,41 +51,121 @@ function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
...
+      state.resourceIndex < manifest?.readingOrder?.length - 1
+    ) {
+      dispatch({ type: 'SET_DATA', data: null });
+      dispatch({ type: 'SET_PAGENUM', pageNum: 1 });
 
-  const goBackward = React.useCallback(() => {
-    console.log('unimplemented');
-  }, []);
+      const data = await loadResource(
+        manifest!,
```
kristojorg: Does this need to be force unwrapped even though you are using the `else if` above to make sure it's there?

src/PdfReader/index.tsx
kristojorg: I actually don't see it being used anywhere

kristojorg (commented):
I left a couple comments about how I would restructure the reducer stuff, but feel free to take or ignore

src/PdfReader/index.tsx
hokei: It's a requirement of the file object https://github.com/wojtekmaj/react-pdf.  I should rename this state variable to `file` instead

src/PdfReader/index.tsx
```diff
@@ -1,24 +1,82 @@
...
-import PDFContent from './PdfContent';
 
 type PdfState = ReaderState & {
   type: 'PDF';
-  // some pdf-specific state here, like the instance of the
-  // package you are using if necessary
+  // The PDF information has been fetched
+  fetchSuccess: boolean;
+  // Is the PDF ready for display
+  displayReady: boolean;
```
hokei: Let me know how else I might make these two things clearer.  The main reason to do this is because there's no way to know how many pages a PDF has until it's loaded.  However, if you're going from the first page of Chapter 3 to to the last page of Chapter 2, we don't want it to flash page 1, so I've held off showing anything until we know that you're loading the correct page.

hokei:
@kristojorg Thanks for all the feedback!  I ended up rewriting a lot of this, it looks a lot cleaner now!  Let me know what you think!

hokei:
@kristojorg fixed a small bug but it works well, I think this is ready for sign off

kristojorg (approved):
I haven't been able to see it working because I can't seem to get the proxy thing to work, but the code looks good to me!

PR #26, opened 2021-07-27
Sfr 1202 add color mode

Toxiapo:
This PR adds the color mode function to the web reader. 

We need the ability to switch between three colors while chakra only offers light/dark mode, hence creating the custom hook function which allows for switching between light/dark/sepia colors.


.storybook/preview.jsx
```diff
@@ -12,7 +12,7 @@ export const parameters = {
 
 const withChakra = (StoryFn, _context) => {
   return (
-    <ChakraProvider theme={theme} >
+    <ChakraProvider theme={theme('day')} >
```
kristojorg: Nice, eventually we can make a control in storybook to toggle the color mode. That can be a separate ticket though.

src/ui/PageButton.tsx
```diff
@@ -8,21 +9,27 @@ export type PageButtonProps = {
 
 function PageButton(props: React.PropsWithoutRef<PageButtonProps>) {
   const { children, onClick, ...rest } = props;
+  const mainBgColor = useColorModeValue('ui.white', 'ui.black', 'ui.sepia');
```
kristojorg, line 12: Maybe in black let's make it `gray.700` or something and in sepia mode `white`? Idk what the old version does but that seems good to me

src/ui/PageButton.tsx
```diff
@@ -8,21 +9,27 @@ export type PageButtonProps = {
...
         size={{ sm: '40px', md: '50px' }}
         mr={8}
         ml={6}
         cursor="unset"
-        _hover={{ backgroundColor: 'gray.700', color: 'ui.white' }}
+        _hover={{
+          backgroundColor: 'gray.700',
+          color: 'ui.white',
+          border: 'none',
+        }}
```
kristojorg: This is going to need to adapt to the color mode as well unfortunately

src/ui/HeaderButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+        border="none"
+        bg="transparent"
+        textTransform="uppercase"
+        fontSize={0}
+        color={color}
+        _hover={{ bgColor: 'transparent' }}
+        _active={{ bgColor: 'transparent' }}
+        {...props}
+      />
+    );
```
kristojorg: After thinking about it, I know I told you to not use the `headerButton` variant in the theme and make a separate component... buuut I think actually we should just make this the default button style in the theme. So not a variant, just the default (I think the default is a solid variant so this would be the styles for the solid variant). That way any button anywhere adapts to the color mode, which is how it should work I think

kristojorg: Toggle button and page button can still remain separate because those are more specific one-offs, but it seems like this is basically the default. LMK if you disagree tho. I think at the minimum all the color-adaptive stuff in here should be in the default button styles in the theme

src/ui/hooks/useColorModeValue.tsx
```diff
@@ -0,0 +1,14 @@
+import { useTheme } from '@emotion/react';
+import { getColor } from '../../utils/getColor';
+
+const useColorModeValue = (
+  light: string,
+  dark: string,
+  sepia: string
+): string => {
+  const { currentColorMode } = useTheme() as any; // TODO: find out the type for this
```
kristojorg, line 9: I see, the `useTheme` doesn't know that `currentColorMode` exists on the theme... that's unfortunate but I think this is ok for now in that case.

src/ui/theme/components/button.ts
```diff
@@ -53,108 +27,76 @@ const commonToggleButtonStyle = {
...
-          color: 'ui.black',
-        },
-        _disabled: {
-          bg: 'ui.sepia',
-        },
-      };
+    case 'light':
+      bgColor = 'ui.white';
+      color = 'gray.800';
+      break;
```
kristojorg: Do we need this switch statement if we are using the above `getColor` function? Doesn't that do the same thing?

src/ui/theme/index.ts
```diff
@@ -1,26 +1,29 @@
...
 import colors from './foundations/colors';
 import nyplTheme from '../nypl-base-theme';
+import { ColorMode } from '../../types';
+import { getColor } from '../../utils/getColor';
+import getButtonStyle from './components/button';
 
 /**
  * See Chakra default theme for shape of theme object:
  * https://github.com/chakra-ui/chakra-ui/tree/main/packages/theme
  */
```
kristojorg: Maybe add a comment here explaining this is a function because we need to pass in the colorMode and adapt the theme to fit

kristojorg (changes requested):
This looks fantastic! One thing I notice is that the prev/next buttons in dark mode and sepia mode are a bit funky. I think in dark mode they should be a gray and in sepia maybe white? Not sure what the old version does but they're currently not super visible and the hover values are a little off.

Otherwise, love this! App is looking better and better : )

src/ui/theme/components/button.ts
Toxiapo: The Day/Sepia/Night buttons behave a little different from the rest of the buttons, their colors should be consistent across different color modes (Per old design).
So I was thinking to use the `colorScheme` props to override the default setting(dynamic color). I am not sure if this is a good way to handle this.

Edit: I Will add a theme wrapper to the `ToggleButton` component per our slack conversation.

src/ui/PageButton.tsx
Toxiapo: This style is for the Button wrapper, which I think should match the color of the Iframe. Perhaps you mean to change the color of the Circle?

kristojorg: Oh true yes

src/ui/PageButton.tsx
Toxiapo: Do you have a specific colors in mind for each mode? Or I can just set the current color for all of them for now and we edit it later.

src/ui/HeaderButton.tsx
Toxiapo: Yea, I like the default color-adaptive variant idea. 

I've added solid as the default variant and included some default styles. Please let me know if you think we should add more CSS properties to the default.

.storybook/preview.jsx
Toxiapo: Ticket Created.

src/ui/PageButton.tsx
kristojorg: Honestly the button doesn't even need a hover style so we can just make it the same as the base style in each color mode

kristojorg (approved):
LGTM!

PR #27, opened 2021-08-03
update WebpubManifest types

Toxiapo:
This PR updates the WebpubManifest Types with our most update-to-date definitions.

### DIFF
--- a/src/useWebReader.tsx
+++ b/src/useWebReader.tsx
@@ -1,22 +1,21 @@
-import {
-  ActiveReader,
-  AnyConformsTo,
-  AxisNowEpubConformsTo,
-  LoadingReader,
-  WebpubManifest,
-  WebpubPdfConformsTo,
-} from './types';
 import React from 'react';
 import { fetchJson } from './utils/fetch';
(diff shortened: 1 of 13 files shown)


src/useWebReader.tsx
```diff
@@ -1,22 +1,21 @@
...
-  WebpubManifest,
-  WebpubPdfConformsTo,
-} from './types';
 import React from 'react';
 import { fetchJson } from './utils/fetch';
 import HtmlReaderContent from './HtmlReader/HtmlReaderContent';
 import usePdfReader from './PdfReader';
 import useHtmlReader from './HtmlReader';
+import { ActiveReader, LoadingReader } from './types';
+import { WebpubManifest } from './WebpubManifestTypes/WebpubManifest';
```
kristojorg: I would re-export `WebpubManifest` from `types/index.ts` so that we can import it from `./types` here instead of this more complicated import path

kristojorg (approved):
Looks good to me. One minor thing I would change but it's not necessary. Up to you

PR #29, opened 2021-08-04
SFR-1230 fix epub overflow

kristojorg:
This fixes the epub overflow issue by making the header sticky and page buttons fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,13 +192,15 @@ The example will rebundle on change, but you have to refresh your browser to see
 
 ### CORS Proxy
 
-We sometimes run in to CORS errors, and have a system to allow urls in a `WebpubManifest` to be proxied. The example app can take a `CORS_PROXY_URL` environment variable to configure where that proxy should be. For development purposes, there is a simple express-based CORS proxy at `example/cors-proxy.js`. You can run it with `npm run cors-proxy` and then make a `.env` file in the root of the project with the following contents:
+We sometimes run in to CORS errors, and have a system to allow urls in a `WebpubManifest` to be proxied. This is done by passing a `proxyUrl` to the `<WebReader>` component. In order to do that, you must have a proxy running somewhere. 
 
-```
-CORS_PROXY_URL="http://localhost:3001/?requestUrl="
-```
+#### CORS Proxy in the Example App
 
-You will need to run the proxy and the example app in separate terminal sessions simultaneously. With this set up, you'll be able to request resources that have strict CORS settings.
+I have set up a small express-based CORS proxy that can be run for local development. 
+
+- Run the proxy with `npm run cors-proxy`.
+- Pass the proxy url to the example app by setting the following env var in a `.env` file at the root of the project: `CORS_PROXY_URL="http://localhost:3001/?requestUrl="`.
+- In a separate terminal session, start the example app: `npm run example`. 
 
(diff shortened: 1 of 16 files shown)


src/HtmlReader/HtmlReaderContent.tsx
```diff
@@ -2,13 +2,11 @@ import * as React from 'react';
...
 const HtmlReaderContent = () => {
   return (
-    <div style={{ height: '100%', overflow: 'hidden', width: '100%' }}>
-      <div id="D2Reader-Container">
-        <main style={{ height: '100vh' }} tabIndex={-1} id="iframe-wrapper">
-          <div id="reader-loading" className="loading"></div>
-          <div id="reader-error" className="error"></div>
-        </main>
-      </div>
+    <div id="D2Reader-Container">
```
Toxiapo, line 6: Is this being added because there are styles coming from the r2 library?

Toxiapo (approved):

hokei:
Screen Shot 2021-08-04 at 12 58 53 PM
the left-hand button is disappearing on pdfs

hokei:
Resizing the app from wider to narrower seems to mess with the viewports
Screen Shot 2021-08-04 at 1 00 30 PM

hokei:
I wonder if the positioning is messing with the `<SizeMe>` tool that's being used to enforce full-width PDF pages

kristojorg:
@hokei I've updated this to add both a dev proxy (`npm run cors-proxy`) and fixed the styles with PDFs

README.md
```diff
@@ -190,6 +190,16 @@ npm run example
 
 The example will rebundle on change, but you have to refresh your browser to see changes (no hot reloading currently).
 
+### CORS Proxy
```
hokei, line 193: can we specify that this is only for running the example?  eg: CORS_PROXY_URL env variable is useless when the webreader is being used by a different app

src/HtmlReader/HtmlReaderContent.tsx
hokei: This is necessary because the D2reader looks specifically for the object with this id to inject the reader

https://github.com/d-i-t-a/R2D2BC/blob/8543345bcd91f1ae740b7910fa9ca8bed7717aca/src/index.ts#L568
(When `iframeNavigator.start` is called, the rest of the reader divs get created and injected into `mainElement`)

hokei:
I think we  need to set the proxy in the env variable in our vercel deploy

hokei (changes requested):
sorry forgot to hit "request changes"

src/HtmlReader/HtmlReaderContent.tsx
Toxiapo: Got it, thanks!

kristojorg:
Ok I updated the readme

hokei (approved):

Toxiapo (changes requested):
sorry, one more thing. The page button background on sepia/dark mode does not match.

kristojorg:
Good catch @Toxiapo. I fixed that. There are still scroll bars sometimes when there shouldn't be, but I believe it is an issue in R2D2BC not in our code at this point and we should fix it later.

Toxiapo (approved):

PR #31, opened 2021-08-05
Sfr 1249 unstyled toc

Toxiapo:
This PR adds the unstyled TOC element to the page with navigation functionality.


src/HtmlReader/index.tsx
```diff
@@ -73,6 +82,7 @@ export default function useHtmlReader(args: ReaderArguments): ReaderReturn {
     fontSize: 16,
     fontFamily: 'sans-serif',
     reader: undefined,
+    currentTocUrl: '',
```
kristojorg: If a blank string is meant to indicate that nothing is set, I would prefer to use `null` in general, as this is more specific and will help avoid bugs down the line

src/ui/TableOfContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+            left: 0,
+            width: '100vw',
+            height: '100vh',
+            background: 'lightgray',
+            padding: '1rem 3rem',
+            margin: 0,
+            overflow: 'auto',
+            zIndex: 999999,
+          }}
+        >
```
kristojorg: For base elements, you can use a Chakra `Box` and the `as` prop to change the element below: `<Box as="ul">` which allows you to use the theme tokens. Don't need to change it here tho until we do the styling, but thought you should know!

src/ui/TableOfContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+          {manifest.toc.map((content) => (
+            <li key={content.title}>
+              <Link
+                href={content.href}
+                onClick={tocLinkHandler}
+                fontSize={`${
+                  readerState?.currentTocUrl === content.href ? 3 : 2
+                }`}
```
kristojorg: I don't think you need the string interpolation here, you can just pass a number to the `fontSize`

src/HtmlReader/index.tsx
```diff
@@ -20,7 +21,8 @@ export type HtmlAction =
   | { type: 'SET_COLOR_MODE'; mode: ColorMode }
   | { type: 'SET_SCROLL'; isScrolling: boolean }
   | { type: 'SET_FONT_SIZE'; size: number }
-  | { type: 'SET_FONT_FAMILY'; family: FontFamily };
+  | { type: 'SET_FONT_FAMILY'; family: FontFamily }
+  | { type: 'SET_CURRENT_TOC_URL'; currentTocUrl: string };
```
kristojorg, line 26: If the `goTo` function takes a `Locator`, does it make sense to just store the `currentLocation` in the state? The problem I guess would be keeping that up to date...

kristojorg (approved):
I think there should be some tests for clicking on TOC elements and making sure you do get navigated. You can write those now or when you do the styling though.

In general this looks good, the only thing that worries me is the `currentTocUrl` because we will have to make sure that is kept up to date all the time. I don't actually know that right now we need to be showing the "current position" in the TOC. But this does appear to work, so I guess let's keep it for now. Ultimately I think it would be better to just store the `currentLocation` in the state, but I'm not sure how difficult that would be.

src/ui/TableOfContent.tsx
Toxiapo: Thank you!!

src/HtmlReader/index.tsx
Toxiapo: I see, I do also need to update the goForward and goNext function, I was wondering if it's good to expose the D2 object to the component, which should contain the updated positions.

kristojorg: What do you mean expose the D2 object to the component?

Toxiapo: Pass down the D2(reader) object to the TOC component so we can get the position directly instead of adding the extra reducer.

Toxiapo:
@kristojorg test added.

src/HtmlReader/index.tsx
kristojorg: We shouldn't pass down the D2Reader because the UI components shouldn't know about it. We should just pass down the `currentLocation`, so in general I think this is ok

src/ui/Header.tsx
```diff
@@ -28,7 +30,7 @@ export default function Header(props: HeaderProps) {
       left={0}
       right={0}
       height={`${HEADER_HEIGHT}px`}
-      zIndex="sticky"
+      zIndex="popover"
```
kristojorg: I'm not sure this should be done, since the header really is "sticky" and not a popover. I think if there are z-index issues I'd rather fix them some other way

tests/TableOfContent.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+  );
+});
+
+test('render Table Of Content', () => {
+  // We need to open the TOC element for TOC links to show up
+  const toggleBtn = screen.getByRole('button', { name: 'Table of Contents' });
+  fireEvent.click(toggleBtn);
+
+  const tocLinkElm = screen.getByRole('link', { name: 'Chapter 1' });
+  expect(tocLinkElm).toBeInTheDocument();
```
kristojorg, line 29: I would add an assertion that the "Chapter 1" is not there before you click the toggle button

kristojorg (changes requested):
Nice work on the tests. It looks good. One small comment about the css styling and one small one about a test. We're basically there though

src/ui/Header.tsx
Toxiapo: My intention was to make sure the TOC popover is always placed at the top(currently resides under the Header Component) instead of the PageButton component. ~~Should I update the page button instead?~~

Edit: Maybe we can ignore this until we add in the styles? We can wrap the popover in a [Portal](https://chakra-ui.com/docs/overlay/menu#rendering-menu-in-a-portal) and that should fix the z-index issue.

kristojorg: Yea I think that's probably the right move. We could also make the TOC popover a higher z index and the header one can stay sticky I think. But yes, for the styles PR

kristojorg (approved):
LGTM

PR #32, opened 2021-08-10
SFR-1226 scroll page

hokei:
Adds pagination and scrolling mode, and makes  the PDF fill page width


src/PdfReader/index.tsx
```diff
@@ -285,29 +338,59 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
+            zIndex="sticky"
+          >
+            <HStack ml="auto" spacing={1}>
+              <Button onClick={() => previousChapter()}>
+                Previous Chapter
+              </Button>
+              <Button onClick={() => nextChapter()}>Next Chapter</Button>
+            </HStack>
+          </Flex>
+        )}
```
kristojorg: Oh interesting approach to add them here. These should be outside of this component though in the manager UI with the page buttons etc. You'd then add a `previousChapter` and `nextChapter` (or `previousResource` more accurately) api to the `navigator`. I think we should decide if we are going to have these buttons though first

kristojorg (changes requested):
Some notes that are only marginally related to this PR:

1. First of all a note that I do notice some pretty significant delay when loading the PDF at /pdf. I don't know if this is because it is 5MB or because of the proxy. We should probably make a ticket to add a progress indicator when loading. From what I see in the console, it looks like the proxy is maybe downloading the full PDF, then sending it over. I wonder if instead it can be streamed? 
1. In paginated mode, I think that the PDF should probably not overflow the height of the page so there is not scrolling whatsoever, similar to how an epub works. That can be a separate ticket (or you can disagree) though.
1. I think we should add a background color to the page around the PDF because there does seem to be somewhat common blank pages in PDFs, and I think it's important to show the user that they are seeing the actual page, it's just blank. I think a background color would make it obvious that you're viewing an actual page. In scrolling mode, you would see the pages separated, just be able to scroll through them. Maybe I'm crazy, but since PDFs don't reflow automatically and are actually page-based things, this seems like it would be a clearer UX. It's also how I'm used to seeing PDFs. Then again, I do like how similar the PDF setup looks to the EPUB setup right now!
1. We decided previously not to have the Prev/Next chapter buttons. My personal preference would be to use the same forward/back button as is used in paginated mode, but it navigates you resources in scrolling mode. 

Code looks good other than these notes

hokei:
> First of all a note that I do notice some pretty significant delay when loading the PDF at /pdf.

Yes, let's make a new ticket.  
If it's saying "PDF is loading" then that means it's the loaded PDF -> Render process that's taking a long time.  There are tools to handle that, that we should add.  

If it's not saying anything then it is downloading the PDF.  

Importantly, the new ticket should allow us to prioritize.  Library users are in general pretty used to PDFs taking a long time to load, so it may not be launch-blocking for DRB.

> In paginated mode, I think that the PDF should probably not overflow the height of the page so there is not scrolling. 

Yes, and I'll make another ticket for this also.  
I spent some time trying to make it work - the trickiness comes in the fact that if you pass a width and a height in the pdf renderer, it only uses width.  So, we would need to figure out how to calculate the height of a page, given a width, and use the height if it's smaller than container height, and width if it's smaller than container width.   Width/height has to be passed in as props (can't be CSS) so that it's rendered at the correct resolution.  I think the solution here is to find container width/container height and figure out if it's landscape (pass height) or portrait (pass width).  Even when I tried that, there seemed to be some mysterious gutter that appeared, on certain pages of the PDF, and at that point I thought, "new ticket" haha

>  I do like how similar the PDF setup looks to the EPUB setup right now!

Yeah it's a stated goal that we don't want people to be thinking about format when they're reading books, unless they really need to.  I think, 

> I think we should add a background color to the page 

How about if I added a small border to the pages?  That would solve the blank page issue, but still look relatively similar to epubs.  

> We decided previously not to have the Prev/Next chapter buttons. My personal preference would be to use the same forward/back button as is used in paginated mode, but it navigates you resources in scrolling mode.

I separated them because of [this](https://github.com/NYPL-Simplified/web-reader/pull/32#discussion_r686809181)

src/PdfReader/index.tsx
hokei: I thought because the chapter buttons would only appear in PDFs, they should go in a PDF specific place - so that epubs wouldn't be tempted to use them.

src/ui/manager.tsx
```diff
@@ -30,7 +30,9 @@ const WebReaderContent: React.FC<ReaderReturn> = ({
       {navigator && state && (
         <Header readerState={state} navigator={navigator} />
       )}
-      <PageButton onClick={navigator?.goBackward} left={0}>{`<`}</PageButton>
+      {state && !state.isScrolling && (
```
hokei: this is the reason why I don't think we should re-use the page buttons for PDF chapter buttons - the "sensible UI" for epubs is that, in scrolling mode, the page buttons don't appear.

Toxiapo:
> > First of all a note that I do notice some pretty significant delay when loading the PDF at /pdf.
> 
> Yes, let's make a new ticket.
> If it's saying "PDF is loading" then that means it's the loaded PDF -> Render process that's taking a long time. There are tools to handle that, that we should add.

@hokei An UI thing that I noticed was the lags on scroll mode, maybe that's related and could be put on the same ticket. (Try to open the setting button on scroll mode).

src/ui/manager.tsx
kristojorg: Yea I guess in both PDFs and EPUBs no matter which mode you are in, you need a forward and back button though, and to me it seemed sensible to have just one forward/back button. But yea we don't have design guidance so it doesn't really matter. If you want to separate them then we can separate them, but they should be rendered here in the manager, not in the reader content component itself

src/PdfReader/index.tsx
kristojorg: We will need them for EPUBs as well if we choose not to show the forward/back button on scroll mode. EPUBs effectively work the same way as PDFs here - you still have to navigate to the next/prev resource using a button in scroll mode

hokei: ah good point.  I will repurpose these buttons, that makes sense.

hokei:
Created https://jira.nypl.org/browse/SFR-1263 for performance, and added the scrollbar as AC to 
https://jira.nypl.org/browse/SFR-1223

src/PdfReader/index.tsx
```diff
@@ -154,7 +178,7 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
     // do nothing if we haven't parsed the number of pages yet
     if (!state.numPages) return;
 
-    if (state.pageNumber < state.numPages) {
+    if (state.pageNumber < state.numPages && !state.isScrolling) {
```
kristojorg, line 182: A comment would be useful at these if statements

src/PdfReader/index.tsx
```diff
@@ -285,29 +317,39 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
-                  <Page key={`page_${index + 1}`} pageNumber={index + 1} />
-                ))}
-              {!state.isScrolling && (
-                <Page pageNumber={state.pageNumber} loading={<></>} />
-              )}
-            </>
-          )}
-        </Document>
-      </Flex>
+      <>
```
kristojorg, line 326: Do you still need this enclosing fragment?

kristojorg (approved):
Looks good to me, just a couple minor comments

PR #33, opened 2021-08-11
Sfr 1208 add toc style

Toxiapo:
This PR adds `Menu` style to the TOC popover.


src/ui/TableOfContent.tsx
```diff
@@ -1,10 +1,75 @@
...
+  const activeHover = useColorModeValue('ui.white', 'ui.white', 'ui.sepia');
+  const borderColor = useColorModeValue(
+    'ui.gray.medium',
+    'gray.500',
+    'yellow.600'
+  );
+
+  const activeColor = 'ui.black';
+  const activeBgColor = 'ui.gray.light-cool';
+
```
Toxiapo: It might be good to separate this component and made it into a new file?

src/ui/TableOfContent.tsx
```diff
@@ -19,60 +83,57 @@ export default function TableOfContent({
...
+                {content.children &&
+                  content.children.map((subLink) => (
+                    <TocItem
+                      key={subLink.title}
+                      href={subLink.href}
+                      title={subLink.title}
+                      isActive={readerState?.currentTocUrl === subLink.href}
+                      onClickHandler={tocLinkHandler}
+                      pl={10}
+                    />
```
Toxiapo: This assumes we are only dealing with single-level `children`, will there ever be more than one level of children exists in a TOC?

src/ui/TableOfContent.tsx
```diff
@@ -1,10 +1,74 @@
...
-import { Link } from '@chakra-ui/react';
 import Button from './Button';
 import { HEADER_HEIGHT } from './Header';
+import useColorModeValue from './hooks/useColorModeValue';
+
+type TocItemProps = React.ComponentPropsWithoutRef<typeof MenuItem> & {
+  href: string;
+  title: string | undefined;
+  isActive: boolean;
+  onClickHandler: (evt: any) => void;
```
kristojorg: I don't think you need to add this, as you can just pass in an `onClick` right?

src/ui/TableOfContent.tsx
kristojorg: I actually don't know, but I think we can assume we don't need to handle that for now. Aferdita also has never seen it.

kristojorg (changes requested):
Looks great! Just the one unnecessary new type I think : )

src/ui/TableOfContent.tsx
kristojorg: I don't think its necessary unless the file gets too long to navigate nicely. Seems fine to me, especially because it's only used in this file

src/ui/TableOfContent.tsx
Toxiapo: Updated, thanks!

src/ui/TableOfContent.tsx
```diff
@@ -19,60 +77,57 @@ export default function TableOfContent({
...
+            mt="-2px" // Move the popover slightly higher to hide Header border
+            overflow="auto"
+          >
+            {manifest.toc.map((content) => (
+              <React.Fragment key={content.title}>
+                <TocItem
+                  href={content.href}
+                  title={content.title}
+                  isActive={readerState?.currentTocUrl === content.href}
+                  onClick={tocLinkHandler}
```
kristojorg, line 113: I think what I would do is `onClick={() => handleLinkClick(content.href)}`, but this appears to also work. Just thought I'd leave a comment about it

kristojorg (approved):

PR #34, opened 2021-08-13
add fullscreen button and function

Toxiapo:
This PR implements Fullscreen functionality to the page using the [Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API)


kristojorg (approved):
Looks good! If that toggleFullscreen code came from somewhere, a comment with the source would be good : )

Toxiapo:
> Looks good! If that toggleFullscreen code came from somewhere, a comment with the source would be good : )

@kristojorg The functions were from our old reader, I can link the code in a comment, and I defined those types with the help of vscode haha.

kristojorg:
Oh nice, nvm then we don't need to link it : ) This looks good to me let's just update the branch and merge!

PR #37, opened 2021-08-19
Offline Service Worker

kristojorg:
This adds offline functionality via two methods:
1. adding all publication resources to a cache called `"webpub-cache"` in the `usePublicationSW` hook.
2. configuring a service worker to check for resources in that cache on each fetch event, and respond immediately if it detects one (within an expiration window).

The host app is responsible for using this `initWebReaderSW()` function within their own service worker script, and for pre-caching their app shell files (html, js, css). I have implemented this in the example app, so if everything goes well, the vercel-deployed version here should, after a first load of the site, work entirely offline (🤞 )

## React-pdf WebWorker

@hokei I did have to use the generic `react-pdf` import to make this work (it was actually caused by a parcel update), which was going to have to happen anyway since we couldn't use either a parcel or webpack specific import without knowing what our consuming apps would use. We will have to add documentation to tell the host app how to include the react-pdf web-worker themself.

I have manually copied the pdf worker from react-pdf to `example/static/pdf-worker` so we could host it and tell react-pdf about it as their documentation suggested. We will need to figure out how to make this work for host apps.


hokei:
> I have manually copied the pdf worker from react-pdf to example/static/pdf-worker so we could host it and tell react-pdf about it as their documentation suggested. We will need to figure out how to make this work for host apps. 

Looks like we both arrived at this conclusion!  Once we get this in DRB we can get a good sense of what it entails to make this work for host apps and document it then.  No need to worry much about it now.

example/index.tsx
```diff
@@ -12,33 +12,35 @@ import {
...
-    fetch('http://localhost:1234/samples/pdf/degruyter.json').then((res) => {
-      res.text().then((text) => {
-        setValue(text);
-      });
-    });
-  });
-
-  return <div>{value}</div>;
-};
+const pdfProxyUrl = process.env.CORS_PROXY_URL as string | undefined;
```
hokei, line 23: do all consuming apps need an environment variable exactly called CORS_PROXY_URL ?

example/index.tsx
```diff
@@ -12,33 +12,35 @@ import {
...
 
 const App = () => {
+  usePublicationSW([
```
hokei, line 34: For running the example app, I wonder if we should cache some things, but not everything, by default.  The purpose of the example app is to aid in development and I'd hate for something to actually be failing, but it looks like it's working because it's been cached.

localhost-key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
hokei, line 1: missed something - what is this for?

hokei (commented):
Some questions, but this looks really good to me.  Amazing work.  

I'll approve if we agree to cache everything all the time.

example/index.tsx
kristojorg: No, this is just how we define it when running our example. Consuming apps can set it however they want and then simply pass it to the web reader as a prop

example/index.tsx
kristojorg: I disabled the SW in development, but yea not a bad idea. I can just cache one example for offline

localhost-key.pem
kristojorg: This is because in order for the SW to run on localhost, you have to enable HTTPS in development. It's a pain but I put some info in the README about it. You don't need it generally because I disable the SW in development by default... but if you do want to work on the SW for example then you need to run `npm run example:sw` which will use these keys for HTTPS

hokei (approved):
I'll give a pre-emptive approve on this.  I would prefer to not cache everything by default on the dev site, but I'm happy to have that in a different ticket as well.

PR #40, opened 2021-08-23
SFR_1062 Decryption

kristojorg:
This PR is for [SFR-1062](https://jira.nypl.org/browse/SFR-1062) to add decryption of AxisNow content. Basically, you can now pass a `getContent` function to the web reader which will be used to fetch and otherwise manipulate content before it makes its way to the iframe. In the example app, I create a `getContent` function my creating an AxisNow decryptor that also combs through the html content for embedded resources and decrypts those as well.

In the example app, this fetching, decrypting and re-embedding of resources is done in a Web Worker using [Comlink](https://github.com/GoogleChromeLabs/comlink) to communicate. We will basically be copying this logic over to CPW (or otherwise moving it upstream into the Decryptor itself. 

One thing to note: we currently use `createObjectUrl` when re-embedding resources, which is a nice, performant way to get a url for some local blob data so we don't have to pass the full object around all the time. However, we should technically be releasing these object urls after we are done with them (like when switching to the next chapter). We need to make a ticket to figure out how to do that. One way would be keeping a list of all object urls created and clearing them out when we start the process for a new chapter. 

Another thing to note: we currently don't handle urls embedded in css (for example embedded fonts or images). Although, I'm not sure if we need to or if those resources would even be encrypted.

I will need to add a token for vercel to deploy this app with the AxisNow decryptor in order for the example to work.


kristojorg:
Update: the vercel preview is now working. You can select the first "Encrypted EPUB Example" option to view. The second "Manually Decrypted EPUB" is the same book after decryption locally

hokei:
> Although, I'm not sure if we need to or if those resources would even be encrypted. 

It's currently decrypting axis books without, so we can leave it as is.  If we have different encryption in the future that requires us to do this, we can add it then.  Each publisher encrypts its content differently and we can only work with the books we actually have.

README.md
```diff
@@ -181,6 +181,10 @@ This is the folder structure:
 /.storybook         # storybook config
 ```
 
+### Decryption
+
+In some instances, decryption of a resource must be performed upon receiving it. For this purpose you can pass an optional `getContent` function that will be used to fetch (and possibly decrypt) the resources to be displayed. It is suggested to perform heavy tasks such as decryption in a Web Worker. The example app shows an example of this being done using the private NYPL AxisNow decryptor (not available publicly). It sets up a WebWorker and communication channel using [Comlink](https://github.com/GoogleChromeLabs/comlink) to fetch, decrypt, and re-embed encrypted subresources.
```
hokei: can you elaborate on what you mean by 
> decryption of a resource must be performed upon receiving it.

hokei: Does this mean that pretty much all the decryption happens within webreader, under the hood?   Nothing additional for the consuming app to set up?

example/axisnow/decryptAndEmbed.ts
```diff
@@ -0,0 +1,126 @@
+import Decryptor from '@nypl-simplified-packages/axisnow-access-control-web';
```
hokei: We should make sure that any users know that this is a specific decryption scheme, and that it's not supposed to work unless you're working with very specific books.  

(and that to any member of the public that's trying to use this repo, they should ignore this)

hokei: I think some of this file should be extracted to be an interface.  Eg:  a `decryptChapter` and `embedImageAssets` function should be available regardless of which Decryptor you choose?  But I think we want Decryption functionality to be optionally available in the webreader. 

However, it's up to you to actually implement your decryptor

example/index.tsx
```diff
@@ -57,6 +60,9 @@ const App = () => {
               proxyUrl={pdfProxyUrl}
             />
           </Route>
+          <Route path="/axisnow-encrypted">
```
hokei, line 73: Only if this is easy:  Can we get this to only show up if you actually have the axis stuff?  (Or throw a reasonable error message if you don't have it)?

kristojorg:
> It's currently decrypting axis books without, so we can leave it as is. If we have different encryption in the future that requires us to do this, we can add it then. Each publisher encrypts its content differently and we can only work with the books we actually have.

RE urls embedded within css: It is working right now without, but I think that is simply because we don't have (or haven't come across) books that use urls in their css. I don't think this is a policy of axisnow decryption, just a happenstance. If some book had a part where they use `background-image: url("...something")`, I believe it would fail because that resources would presumably be encrypted. However, I agree we can probably ignore this for now.

example/axisnow/decryptAndEmbed.ts
kristojorg: Yea we could do that although I sort of doubt anyone will need client side decryption initially, and if they do idk that our methods will work for them, or if their setup would be significantly different ya know? Like I could make it so the app takes a: `type Decryptor = (in: Uint8Array) => Promise<Uint8Array>`... and then we would call these `decryptChapter` and `embedImageAssets` internally. I'm just dubious whether anyone else's decryptor would necessarily function similarly enough to make that useful. It also wouldn't work on PDF content. So I guess I was more inclined to just say "hey, here's a way to define a function that takes the resourceUrl and returns a string. Do whatever you need to do in there"

README.md
kristojorg: Ok I updated this, lmk what you think

example/index.tsx
kristojorg: Did that here: https://github.com/NYPL-Simplified/web-reader/pull/43

example/axisnow/decryptAndEmbed.ts
hokei: That makes a lot of sense.  Let's leave it as is, until we know exactly what others' use cases are.  DRM is such a ... contrived problem that I wouldn't be surprised if everyone has very specific things they need their drm to solve ...

hokei:
> RE urls embedded within css: It is working right now without, but I think that is simply because we don't have (or haven't come across) books that use urls in their css. I don't think this is a policy of axisnow decryption, just a happenstance. If some book had a part where they use background-image: url("...something"), I believe it would fail because that resources would presumably be encrypted. However, I agree we can probably ignore this for now.

This is really good to know.  A thing I've been pondering for a while, that we can talk about not in this PR, is where can we store information like this that isn't just our heads?

README.md
```diff
@@ -196,6 +196,23 @@ This is the folder structure:
...
+1. If decryption cannot be performed in a streamer, the web-reader can support client-side decryption of licensed content. This is done by passing a `getContent` function to either the `<WebReader>` component or the `useWebReader` hook. It has the type signature `(resourceUrl: string) => Promise<string>`, and can thus be used to fetch and decrypt (or otherwise manipulate) content before it is passed to the iframe for rendering.
+
+The `AxisNow Encrypted EPUB` example shows how this is done using the private NYPL AxisNow decryptor. The AxisNow scheme is a specific DRM technique not publicly available and the repo and code for the decryptor cannot be shared. Thus this example will not work for the public, but you can read the example code to see how we use the private `Decryptor` package to:
```
hokei, line 207: I wonder if this is OK to have in this repo.

hokei:
OK this all looks good to me from a code standpoint.  My outstanding questions are around whether or not this is OK to put in this repo.  Can you merge the /examples branch to this branch, and then ask Leonard for a gut check?  If it looks good then feel free to merge.

README.md
kristojorg: I think so? It doesn't describe anything in particular about how decryption works

kristojorg:
@leonardr I'm making you a reviewer on this PR because @hokei wanted a gut check from you about whether this detail around decryption is allowed in this repo. All of the actual decryption logic still lives in the private repository, I'm just using the interface to perform decryption here. Hopefully it's alright? LMK what you think

example/index.tsx
```diff
@@ -103,4 +155,60 @@ const App = () => {
...
+  const [getContent, setGetContent] = React.useState<null | GetContent>(null);
+  const [error, setError] = React.useState<Error | undefined>(undefined);
+
+  React.useEffect(() => {
+    async function setupDecryptor(): Promise<
+      (href: string) => Promise<string>
+    > {
+      const params = {
+        book_vault_uuid: '6734F7F5-C48F-4A38-9AE5-9DF4ADCFBF0A',
+        isbn: '9781467784870',
```
leonardr, line 185: The unit tests in the circulation manager reveal the existence of "book_vault_uuid" and "isbn", so this is okay, but if this is a real book_vault_uuid it should be changed for a fake one. Real ISBN is fine.

example/static/samples/axisnow/encrypted/manifest.json
```diff
@@ -0,0 +1,676 @@
+{
```
leonardr, line 1: This isn't a big deal since the manifest doesn't contain any copyrighted material, but unless you need something specific from this manifest (like the embedded media) it would be a little better to use the manifest for the public domain title I sent you earlier today.

leonardr:
The only details exposed here are the names of the variables the AxisNow decryptor takes as input, and that was already public through the circulation manager, so I'm OK with this.

hokei (approved):
This looks good to me - just the comment about making sure the book_vault_id is a fake one.  

At this point let's prioritize merging this in and if we need to make changes in the future we can do that as bugfixes.  If you want me to take another look after resolving merge conflicts and stuff let me know!

PR #41, opened 2021-08-23
Sfr 1262 setup cypress test

Toxiapo:
This PR adds tests using the cypress library.

**Includes**:
- Initial renders of the app and make sure all necessary components are attached.
- Is able to navigate between pages using the TOC links and the next or previous button.

One thing I was having trouble with is getting the new contents after the iframe reload(change page).  It seems like the commands are executing in an _asynchronous_ manner, hence the `cy.wait(100)` lines on the code wherever I expecting a new `iframe` should be returned, without the wait, the 'iframe' would have the reference of the old content.


cypress/integration/navigations.ts
```diff
@@ -0,0 +1,58 @@
...
+    // Open TOC menu
+    cy.findByRole('button', { name: 'Table of Contents' }).click();
+
+    // Open chapter 1
+    cy.findByRole('menuitem', { name: 'I: Down the Rab­bit-Hole' }).click();
+    cy.wait(100); // wait for the nested iframe to render
+
+    cy.iframe(IFRAME_SELECTOR)
+      .find('.subtitle')
+      .contains('Down the Rab­bit-Hole');
```
Toxiapo: I had to explicitly set a wait time for the click event to get registered/ or given some time for the iframe to load the new contents.

kristojorg: I think that's to be expected here because we don't get a promise back from R2D2BC letting us know when this is complete, so I think setting a timer is all we can do, unless cypress has a wait to just wait for something to be true (with a given timeout)

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,13 @@
...
+# jobs:
+#   cypress:
+#     if: github.event_name == 'deployment_status' && github.event.deployment_status.state == 'success'
+#     runs-on: ubuntu-20.04
+#     steps:
+#   - name: Dump GitHub context
+#     env:
+#       GITHUB_CONTEXT: ${{ toJson(github) }}
+#     run: |
+#       echo "$GITHUB_CONTEXT"
```
kristojorg: Is this meant to be commented out?

cypress.json
```diff
@@ -0,0 +1,4 @@
+{
+  "baseUrl": "http://localhost:1234",
+  "chromeWebSecurity": false
```
kristojorg: Just curious what this does?

cypress/integration/navigations.ts
```diff
@@ -0,0 +1,58 @@
+import { IFRAME_SELECTOR } from '../support/constants';
+
+describe('navigations', () => {
```
kristojorg: I would give this a more descriptive name like "navigating an EPUB" or something

cypress/integration/navigations.ts
kristojorg: I would personally add this same selector but asserting it does not exist before you navigate. That way we are sure it's not passing simply because we are in scrolling mode and that subtitle appears further down the page.

cypress/integration/navigations.ts
```diff
@@ -0,0 +1,58 @@
...
+    // Open chapter 1
+    cy.findByRole('menuitem', { name: 'I: Down the Rab­bit-Hole' }).click();
+    cy.wait(100); // wait for the nested iframe to render
+
+    cy.iframe(IFRAME_SELECTOR)
+      .find('.subtitle')
+      .contains('Down the Rab­bit-Hole');
+  });
+
+  it('next and previous button', () => {
```
kristojorg: It can be useful to use the "it" to form a sentence for the test names, like `it("should navigate forward and backwards with page buttons")`. Although it's not always possible to phrase it like that, but when we can I think we should try to

cypress/integration/navigations.ts
```diff
@@ -0,0 +1,58 @@
...
+    // load iframe
+    cy.iframe(IFRAME_SELECTOR);
+
+    cy.findByRole('button', { name: '>' }).click();
```
kristojorg: This is actually indicating that we have an A11y bug: we need to give that button a better aria-label: "Next Page" manually. By default buttons use their content as their aria label, but in this case it doesn't make much sense for screen readers, so we should add an explicit one. You can make a ticket for that in the staging sprint and put a `FIXME` comment with the ticket number here

cypress/integration/navigations.ts
kristojorg: Although maybe this strategy would be useful: https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress/#slow-loading-frames

cypress/integration/navigations.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    cy.wait(100);
+    cy.iframe(IFRAME_SELECTOR)
+      .findByRole('img', {
```
kristojorg: I would also use this strategy to make a custom command to get the iframe: https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress/#custom-command

cypress/integration/render-app.ts
```diff
@@ -0,0 +1,22 @@
+import { IFRAME_SELECTOR } from '../support/constants';
+
+describe('render app', () => {
```
kristojorg: I think I would make this specific to a certain format. Like have:

- HTML Reader
  - Renders manager controls (this test)
  - Renders loading page while loading
  - Navigation (the tests in integration/navigations)
  - Display settings (for font family, size, etc)
- PDF Reader
  - ... same set of tests basically

kristojorg (commented):
This is a great start! Left a few comments I hope are helpful? I don't think you need to make all those tests in this pr, especially if this PR is just about getting cypress et up and running. So let me know what the scope of this PR is if you want me to wait on my comments for stuff that's outside the scope (like organization of the tests).

However, for this PR I do think you should add a little README testing section talking about how to run these

Toxiapo:
@kristojorg Yep, this PR is to get cypress setup and add the navigation test. We should have some follow-up tickets/subtasks for more test coverage.

I will address the comment update and also add a test section to the readme for this PR, thank you!

cypress/integration/navigations.ts
Toxiapo: I was actually using [this](https://gitlab.com/kgroat/cypress-iframe) library to get the iframe element, but I think I can remove it and create the custom command from the blog since our use case is simple.

.github/workflows/cypress.yml
Toxiapo: Yea I haven't set this up yet, this is a file template that @samanthaandrews help me set up to get me started.

cypress.json
Toxiapo: I think this bypasses the **cross-origin** errors and lets you load your page in an iframe.

We don't need it tho. It must be a `debugging` code I wrote and forgot to delete. I will remove it.

cypress/integration/navigations.ts
Toxiapo: Yea that makes sense!

cypress/integration/render-app.ts
Toxiapo: I've added a folder for the HTML reader tests, and also created a ticket to cover the rest of the tests.

kristojorg:
LMK when this is ready to be reviewed again! (or also can move it to a draft until you want me to come back to it)

cypress/integration/navigations.ts
Toxiapo: This did not work :( Even with a long timeout.

Toxiapo:
@kristojorg Yes, it's ready for review! Although I am not sure if we need more setups for the cypress test action, like reporting failed screenshots/videos

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
hokei: is this being used?

Toxiapo: Oh, it's not! I will delete this.

hokei:
Otherwise this looks good to me and I'll leave it to Kristo for signoff

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/checkout@v2
+
+      - name: Install Cypress
+        run: npm i cypress
+
+      - name: Run Default (Electron)
+        uses: cypress-io/github-action@v2
+        with:
+          install: false
+          config: baseUrl=${{ github.event.deployment_status.target_url }}
```
samanthaandrews: Instead of adding the baseUrl to this config object, let's put it in the `env` object like it is in the CPW repo. Adding it this way means that the `baseUrl` value from the cypress.json file will be overwritten in the CI with this `CYPRESS_BASE_URL` value (this is a fancy Cypress thing). Should look like this:

```
      - name: Run Default (Electron)
        uses: cypress-io/github-action@v2
        with:
          install: false
        env:
          CYPRESS_BASE_URL: https://qa-web.openebooks.us
          # Recommended: pass the GitHub token lets this action correctly
          # determine the unique run id necessary to re-run the checks
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

File: .gitignore [hunk 11]
```diff
@@ -16,3 +16,7 @@ storybook-static
 
 # the parcel built example app
 www
+
+# cypress
+cypress/screenshots
+cypress/videos
```
samanthaandrews: nit: add end of line (eol) padding here

samanthaandrews (changes requested):
With just those very small changes I'm ok with merging this tomorrow! Looks great!

.github/workflows/cypress.yml
Toxiapo: For some reason, I was getting a type error (Expect String | Number | Boolean) if I set it as a .env value. 
(Line 38, is this what you mean?)

samanthaandrews: Yes... hmm. I'll look at it tomorrow and see if I can figure out why it wouldn't like that value!!

samanthaandrews (approved):
I ran the tests locally and they're working for me 🎉 nice work

.github/workflows/cypress.yml
Toxiapo: ok. it works! It was my vscode linter that was complaining! Thanks!

PR #42, opened 2021-08-23
SFR-1223 Zoom

hokei:
Implemented Zoom and made the contents fit the container * 

*There's still a bug where if you change the height to be smaller, somehow `useMeasure` does not seem to update.  I really would love to extract this into another bug because I've spent way too long on this already 

Also, this seems to have really slowed down the app and any insight into why, would be helpful. 

Annotations also don't line up in the pdf collection example, I do think I need to fix this before merge.


src/PdfReader/index.tsx
```diff
@@ -20,8 +20,15 @@ type PdfState = ReaderState & {
...
   // resource once it is parsed
   pageNumber: number;
+  scale: number;
+  pdfHeight: number;
+  pdfWidth: number;
+  pageHeight: number | undefined;
+  pageWidth: number | undefined;
 };
 
+pdfjs.GlobalWorkerOptions.workerSrc = `pdf.worker.min.js`;
```
hokei, line 29: this is necessary to get the worker functional.  I'm not sure if there's a more elegant way to do this

src/ui/manager.tsx
```diff
@@ -36,10 +36,10 @@ const WebReaderContent: React.FC<ReaderReturn> = ({
         width="100vw"
         bg={bgColor}
         // accounting for the prev/next buttons
-        px={{ sm: 10, md: '5vw' }}
+        // px={{ sm: 10, md: '5vw' }}
```
hokei: TODO:  I think I can revert this.

src/utils/throttle.ts
```diff
@@ -0,0 +1,13 @@
+export default function throttle(
```
hokei: TODO: I think this isn't used anymore

src/PdfReader/useMeasure.tsx
```diff
@@ -0,0 +1,68 @@
...
+declare const window: Window &
+  typeof globalThis & {
+    ResizeObserver: any;
+  };
+export type Dimensions = Pick<
+  DOMRectReadOnly,
+  'x' | 'y' | 'top' | 'left' | 'right' | 'bottom' | 'height' | 'width'
+>;
+export type UseMeasureRef<E extends Element = Element> = (element: E) => void;
+export type UseMeasureResult<E extends Element = Element> = [
```
hokei, line 12: Seems like this is causing a lint error in the import of React - something to do with the react version?

src/PdfReader/useMeasure.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  React.useLayoutEffect(() => {
+    if (!element) return;
+    observer.observe(element);
+    return () => {
+      observer.disconnect();
+    };
+  }, [element, observer]);
+
+  return [ref, rect];
```
hokei, line 67: There's also a typescript error here that cascades into a type error here: https://github.com/NYPL-Simplified/web-reader/pull/42/files#diff-8c9d1a11768b098f16811aec65d3a9bc7c4baad23d0f7c13fea2150d1cf74d2dR433

I'd love some thoughts on the type issues that happen here, because I'm not sure I understand all of the generics going on, and would love to know how one would debug this.

src/PdfReader/index.tsx
```diff
@@ -1,6 +1,6 @@
-import { Document, Page } from 'react-pdf/dist/esm/entry.parcel';
-
-import React from 'react';
+import { Document, Page, PageProps } from 'react-pdf';
+import { pdfjs } from 'react-pdf';
```
kristojorg: Are we meant to have two separate imports from react-pdf here?

src/PdfReader/index.tsx
kristojorg: I'm hoping we don't have to actually copy the file from there to our repo... but we can do this for now

src/PdfReader/index.tsx
```diff
@@ -174,6 +211,41 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
+    },
+    []
+  );
+
+  //TODO: Somehow, this window size updates when height
+  React.useEffect(() => {
+    if (containerSize) {
+      resizePage(state.pdfWidth, state.pdfHeight, containerSize);
+    }
+  }, [containerSize, state.pdfWidth, state.pdfHeight, resizePage]);
```
kristojorg, line 246: I don't think you need to define the `resizePage` outside the `useEffect`, I think it can just be written inline there, but I may be wrong?

src/PdfReader/index.tsx
```diff
@@ -319,43 +401,63 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
     });
   }
 
+  function onRenderSuccess(page: PageProps) {
+    if (!page.height || !page.width || !containerSize) return;
```
kristojorg: Should we maybe throw an error in this case? Otherwise they'll just have a loading screen forever right? Idk if this is a case that might actually happen or not tho

src/PdfReader/index.tsx
kristojorg: Nvm I see it is also used below

kristojorg (commented):
Ok my thoughts so far. I tried merging main into this and then got parcel errors, but I did find the solution to type errors you were seeing. Zooming looks great in the preview deploy. One thing to note for the future is when we change the buttons to be zoom specific, we should probably have a "100%" button to set you back to baseline (we might also want this in the font size version, come to think of it). I also wonder how the browser zoom should interact with this, from an a11y perspective. Should we zoom the PDF when you press "command +"?

kristojorg:
Ok my parcel issues were because I checked out the `SFR-1223_zoom` branch and not `SFR-1223_zoom_3` haha I'm a dummy.

hokei:
@kristojorg this should be good to go, but I'm not sure why the size is failing when I don't think it actually exceeded the size limit?

kristojorg:
I'm not sure why it was reported that way but I just increased the bundle size

kristojorg (approved):
LGTM!

PR #43, opened 2021-08-25
Clean up examples

kristojorg:
This PR to my decryption branch cleans up the examples and adds an error message for when the AxisNow decryptor is not available, which will be the case for most public use of the example app.

### DIFF
--- a/example/axisnow/decryptAndEmbed.ts
+++ b/example/axisnow/decryptAndEmbed.ts
@@ -1,5 +1,3 @@
-import Decryptor from '@nypl-simplified-packages/axisnow-access-control-web';
-
 type DecryptChapter = (resourceUrl: string) => Promise<string>;
 
 type DecryptorParams = {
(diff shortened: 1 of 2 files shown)


example/axisnow/decryptAndEmbed.ts
```diff
@@ -1,17 +1,35 @@
-import Decryptor from '@nypl-simplified-packages/axisnow-access-control-web';
-
 type DecryptChapter = (resourceUrl: string) => Promise<string>;
 
 type DecryptorParams = {
   isbn: string;
   book_vault_uuid: string;
 };
 
+type IDecryptorInstance = {
```
hokei, line 8: I still don't understand why this goes under /example/ rather than in /src/  Wouldn't all Decryptors, be they AxisNow or not, share the same interface?

example/index.tsx
```diff
@@ -73,31 +75,75 @@ const App = () => {
...
+              <Text fontSize="sm">
+                These examples are specific to NYPL, and may not work properly
+                without access to private packages.
+              </Text>
+              <UnorderedList p={4}>
+                <ListItem>
+                  <Link to="/axisnow-encrypted">AxisNow Encrypted EPUB</Link>
+                </ListItem>
+                <ListItem>
+                  <Link to="/axisnow-decrypted">Decrypted AxisNow EPUB</Link>
```
hokei, line 133: This book is not in the public domain.  We should find an AxisNow encrypted public domain book -- if we can.

hokei: Or, we can remove the "Decrypted" example for now.  I think the "Encrypted" example might be OK to keep because non-NYPL people will not be able to access it.

kristojorg: Hmm ok yea I'll ask leonard

example/axisnow/decryptAndEmbed.ts
kristojorg: Not necessarily... basically we are letting people pass in a function to handle `fetch` requests, which allows them to do decryption or anything else they need to do, but it doesn't have any opinion about what goes in there. It just takes a `url` and returns a `Promise<string>` where the string is the html content. I did this to provide more flexibility around what people might need to do, or how their decryption works. Maybe their decryptor takes a `string` and not a `Uint8Array`, or maybe they need to pass specific headers in their `fetch` call or something. This allows them to do that however they see fit. We don't specify that decryption has to work any specific way.

hokei:
I think this is good to merge into feature/decryption.

hokei:
Thank you for making this as a separate PR, it's super helpful.

PR #46, opened 2021-08-26
Settings card modular

Toxiapo:
This PRs adds in the functionality to render Settings Card settings based on the rendered media type. Currently added for PDF and HTML!

- Defined types for PDF and HTML reader state.
- Created child components for each type.
- Added tests.


src/types.ts
```diff
@@ -23,17 +23,25 @@ export type Navigator = {
   goToPage: (href: string) => void;
 };
 
-export type ReaderType = 'HTML' | 'PDF';
+//  PDF specific navigator here
+export type PdfNavigator = Navigator;
+
+export type HtmlNavigator = Navigator;
```
samanthaandrews, line 29: Jiayong and I talked about needing to do some work to actually split this type (and the ReaderState) one below, but he noticed that this would require making several changes to pieces of the code that @hokei is working on. I think it would be ok to merge this as it is and create a follow-up issue to split these types.

src/ui/SettingsButton.tsx
```diff
@@ -6,27 +6,184 @@ import {
...
+} from '../types';
 import { Icon, IconNames } from '@nypl/design-system-react-components';
 
 import Button from './Button';
 import ToggleButton from './ToggleButton';
 import ToggleGroup from './ToggleGroup';
 import useColorModeValue from './hooks/useColorModeValue';
 
-export default function SettingsCard({
+const PDFSettings = ({
```
samanthaandrews: My recommendation here is to split `PDFSettings` and `HTMLSettings` components into separate files, just for legibility. This file is getting pretty big with all three components in one.

src/ui/SettingsButton.tsx
```diff
@@ -6,27 +6,184 @@ import {
...
+
+export default function SettingsCard(
+  props: SettingsCardProps
+): React.ReactElement {
   const [isOpen, setIsOpen] = React.useState(false);
   const open = () => setIsOpen(!isOpen);
   const close = () => setIsOpen(false);
 
-  const paginationValue = readerState?.isScrolling ? 'scrolling' : 'paginated';
+  const paginationValue = props.state?.isScrolling ? 'scrolling' : 'paginated';
```
samanthaandrews: There might be an opportunity to destructure props in this component. Something like:

`{ state, type } = props`

and then later in the component you can just say `state?.isScrolling ? ...` and `type === 'PDF'`

src/ui/SettingsButton.tsx
```diff
@@ -6,27 +6,184 @@ import {
...
-export default function SettingsCard({
+const PDFSettings = ({
+  navigator,
+  readerState,
+  paginationValue,
+}: {
+  navigator: Navigator;
+  readerState: PdfReaderState;
+  paginationValue: string;
+}): React.ReactElement => {
```
samanthaandrews: Consider using object destructuring in the PDFSettings & HTMLSettings components with `navigator` and `readerState`. Ultimately I think this is a stylistic preference, so feel free to ignore this suggestion.

samanthaandrews (changes requested):

src/ui/SettingsButton.tsx
Toxiapo: Sounds good! We are destructuring props in the headers component as well so it's good to keep it consistent.

src/PdfReader/index.tsx
```diff
@@ -4,14 +4,15 @@ import {
   ColorMode,
   ReaderArguments,
   ReaderReturn,
-  ReaderState,
   WebpubManifest,
+  PdfReaderState,
 } from '../types';
 import { chakra, Flex, shouldForwardProp } from '@chakra-ui/react';
+import useContainerWidth from '../ui/hooks/useContainerWidth';
```
kristojorg: Did you add this import? It looks like it isn't used

src/types.ts
kristojorg: Sounds like a good plan!

kristojorg (approved):
This looks great! Can we make a follow up ticket to actually split up the navigators and states? 

Also, one thing to keep in mind for the future: in the tests, it would be helpful to also check and make sure the buttons are expect not to be there are actually not there. Not super critical here but it would add a bit of confidence to the tests (otherwise we could just be rendering every button all the time and the tests would pass)

src/PdfReader/index.tsx
Toxiapo: Removed! I am not sure why I imported that..

Toxiapo:
> Also, one thing to keep in mind for the future: in the tests, it would be helpful to also check and make sure the buttons are expect not to be there are actually not there. Not super critical here but it would add a bit of confidence to the tests (otherwise we could just be rendering every button all the time and the tests would pass)

Yeah, that makes sense. I added some comments on the settings card test file so I don't forget! :)

Toxiapo:
@kristojorg @samanthaandrews I've created ticket SFR-1288 to split up the navigators and states. Should be ready for review!

kristojorg (approved):
Looks good!

PR #49, opened 2021-09-02
Sfr 1276 more cypress test

Toxiapo:
This PR adds the display settings test.

- Updated the old tests with the new aria-label.
- I removed some of the 'wait' functions and have it to check the loading screen instead (loading icon hidden). This should make the test run faster.


cypress/integration/html-reader/navigations.ts
```diff
@@ -45,11 +55,10 @@ describe('navigating an EPUB page', () => {
...
       .should('not.exist');
 
-    cy.wait(1000);
-
-    cy.findByRole('button', { name: '<' }).click();
+    cy.findByRole('button', { name: 'Previous Page' }).click();
+    cy.get('#reader-loading').should('be.visible');
 
-    cy.wait(1000);
+    cy.get('#reader-loading').should('not.be.visible');
```
Toxiapo: I am adding these in few places to replace the wait function, and it seems to be working fine. Would love to know if there are other better ways! The goal is to make sure our iframe is fully loaded before accessing the iframe content.

package.json
```diff
@@ -17,6 +17,8 @@
     "test:ci": "jest --ci --coverage --maxWorkers=2",
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
+    "cypress:run:html": "cypress run --spec './cypress/integration/html-reader/*'",
+    "cypress:run:pdf": "cypress run --spec './cypress/integration/pdf-reader/*'",
```
kristojorg: Is there a reason to split these out or should we just run all of them all the time?

cypress/integration/html-reader/navigations.ts
kristojorg: Yea I'm not totally sure, because I guess the iframe itself it loaded, but the reader is still loading the book content, and we are trying to wait on that completing, is that right? If you take this out, does it not retry the next failing test until it is in fact loaded?

kristojorg (commented):
This looks really good, let's try and chat today about the loading state and wating

package.json
Toxiapo: hmmm, yea we don't need to separate these, it's good to run all the tests. My intention was to separate them and make debugging quicker when writing tests for html/pdf.

cypress/integration/html-reader/navigations.ts
hokei: Testing library has a [WaitFor](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) - is that what you're looking for here?

Toxiapo: I think the `cy.get` is similar to WaitFor, where it also "re-try" until it finds the element or timeout reached, but somehow that didn't work.

samanthaandrews:
@Toxiapo showed me the issues he's having with running these tests in FireFox, and it's my recommendation to create a separate Jira ticket to get these tests running in that browser. I think the tests running in Chrome & Electron give us the confidence we need to move forward with merging this code. FWIW I'm having similar issues in CPW with FireFox & Cypress playing well together and it feels like it could take quite a bit of digging in order to understand where the problems are stemming from.

package.json
```diff
@@ -109,8 +109,6 @@
     "parcel": "^2.0.0-rc.0",
     "parcel-config-precache-manifest": "^0.0.4",
     "lint-staged": "^11.1.2",
-    "parcel": "^2.0.0-rc.0",
-    "parcel-config-precache-manifest": "^0.0.4",
```
kristojorg, line 113: Curious if these were meant to be removed?

kristojorg (changes requested):
Looks good, just the one thing about the removed installs?

package.json
Toxiapo: Yeah, they are duplicates of line 109

kristojorg: Oh kk perfect thanks

kristojorg (approved):
Looks great! Glad to have this testing started : )

PR #50, opened 2021-09-07
Sfr 1248 storybook theme control

Toxiapo:
This PR adds theme control to the storybook UI.

The components that support color mode should have a `currentColorMode` property under the `argTypes` object, this value will get stored inside the context object in the preview.js file, and will get passed to the `<ChakraProvider />` as the prop which is responsible for updating the theme.


stories/Header.stories.tsx
```diff
@@ -12,6 +12,11 @@ const meta: Meta = {
     readerState: {},
   },
   argTypes: {
+    currentColorMode: {
+      options: ['day', 'sepia', 'night'],
+      control: { type: 'radio' },
+      description: 'Select between day / sepia / night mode',
+    },
```
Toxiapo, line 19: I wonder if we should make this a global argTypes so that we don't redefine them every time we create a new story?

.storybook/preview.jsx
```diff
@@ -1,21 +1,23 @@
...
 
 // https://storybook.js.org/docs/react/writing-stories/parameters#global-parameters
 export const parameters = {
   // https://storybook.js.org/docs/react/essentials/actions#automatically-matching-args
   actions: { argTypesRegex: '^on.*' },
+  options: { showPanel: true },
 };
 
 const withChakra = (StoryFn, _context) => {
+  const { args } = _context;
```
kristojorg, line 14: Just fyi I put the _ in names usually when a variable is not being used, so in this case you can remove the _ because it's not being used. But not a big deal.

stories/Header.stories.tsx
kristojorg: Yea seems smart to me

stories/Button.stories.tsx
```diff
@@ -19,15 +21,19 @@ const meta: Meta = {
 
 export default meta;
 
-const Template: Story<ButtonProps> = (args) => <Button {...args} />;
+const Template: Story<ButtonProps> = (args) => (
+  <ChakraProvider theme={getTheme(args.currentColorMode ?? 'day')}>
```
kristojorg, line 25: Do you need to do this if you're doing it in the preview.jsx file?

kristojorg (approved):
Looks good to me, though I wonder if we need to render the chakra provider in each story or if we can just do it in the preview.jsx file? I'll approve though since it's not a big deal and it works

stories/Header.stories.tsx
Toxiapo: Updated.

stories/Button.stories.tsx
Toxiapo: Yea this is not needed anymore. Updated!

kristojorg (approved):
LGTM!

PR #51, opened 2021-09-08
Enabled Table of Contents for Multi Resource PDFs

hokei:
Table of Contents is now clickable for Multi Resource PDFs

Single Resource PDFs still don't have a ToC, this will be tackled in another ticket.


src/PdfReader/index.tsx
```diff
@@ -121,14 +122,21 @@ function pdfReducer(state: PdfState, action: PdfReaderAction): PdfState {
...
-) => {
+const getResourceUrl = (
+  index: number,
+  readingOrder: ReadiumLink[] | undefined
+): string => {
+  if (!readingOrder || !readingOrder.length) {
+    throw new Error('A manifest has been returned, but has no reading order');
+  }
+
+  // If it has no children, return the link href
```
kristojorg, line 136: Are you checking if it has children here?

src/PdfReader/index.tsx
```diff
@@ -299,7 +309,8 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
         shouldNavigateToEnd: !state.isScrolling,
       });
 
-      const data = await loadResource(manifest, nextIndex, proxyUrl);
+      const resourceUrl = getResourceUrl(nextIndex, manifest.readingOrder);
```
kristojorg: Shuld `getResourceUrl` just be called inside `loadResource` if they are always called together?

src/PdfReader/index.tsx
```diff
@@ -353,9 +364,31 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
+      };
+
+      dispatch({
+        type: 'SET_CURRENT_RESOURCE',
+        index: getIndexFromHref(href),
+        shouldNavigateToEnd: false,
+      });
+      const data = await loadResource(href, proxyUrl);
+      dispatch({
+        type: 'RESOURCE_FETCH_SUCCESS',
```
kristojorg: I wonder if we should just put this into a `useEffect` hook on its own so that we don't have to manually load the resource in every place we perform navigation. We just set the index in state and the `useEffect` hook runs anytime that state changes to load the resource? I can make a PR with my suggestion

src/PdfReader/index.tsx
hokei: nope I moved this functionality so this is a moot comment

kristojorg (approved):
Looks good! 🎉

PR #52, opened 2021-09-09
Add icons to the PageButton component

Toxiapo:
This PR updates the current PageButton component to render an Icon.

- Added DS icon to the PageButton component
- Updated storybook
- Updated Jest test

### DIFF
--- a/src/ui/Header.tsx
+++ b/src/ui/Header.tsx
@@ -1,6 +1,10 @@
 import React from 'react';
-import { Flex, Link, HStack, Text } from '@chakra-ui/react';
-import { Icon, IconNames } from '@nypl/design-system-react-components';
+import { Flex, Link, HStack, Text, chakra } from '@chakra-ui/react';
+import {
+  Icon,
+  IconNames,
+  LogoNames,
+} from '@nypl/design-system-react-components';
 import { ActiveReader } from '../types';
 import useColorModeValue from '../ui/hooks/useColorModeValue';
 import { toggleFullScreen } from '../utils/toggleFullScreen';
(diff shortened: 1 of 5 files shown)


tests/PageButton.test.tsx
```diff
@@ -4,8 +4,12 @@ import PageButton from '../src/ui/PageButton';
 
 test('render PageButton', () => {
   const onClickHandler = jest.fn();
-  render(<PageButton onClick={onClickHandler}>Page</PageButton>);
-  const pageBtn = screen.getByRole('button', { name: 'Page' });
+  render(
+    <PageButton onClick={onClickHandler} aria-label="Next Page">
+      {'>'}
```
hokei: Optional: You can mount an icon and assert that it's there, if you want? 

Does this check for "Previous page" as well?

src/ui/Header.tsx
```diff
@@ -18,6 +22,14 @@ export default function Header(props: HeaderProps): React.ReactElement {
   const { headerLeft, state, navigator, manifest } = props;
   const linkColor = useColorModeValue('gray.700', 'gray.100', 'gray.700');
   const mainBgColor = useColorModeValue('ui.white', 'ui.black', 'ui.sepia');
+
+  // had to wrap this because I was having trouble adding custom styles to the NYPL Icon
```
hokei: Nit: Code style wise, makes more sense to just go 
`// Add custom styles to the NYPL Icon`

(If you want to mention it to the PR reviewers, you can do it in the PR)

hokei (changes requested):
Looks good to me, the stories file should be updated to show an icon

kristojorg (approved):
Looks good to me! I think in general we don't need to have both jest and cypress tests, so I would probably stick to just writing cypress tests for things unless there is some logic-heavy piece that we would like to write unit tests for. @hokei do you agree?

src/ui/Header.tsx
Toxiapo: got it, thanks!

tests/PageButton.test.tsx
Toxiapo: Added!

hokei (approved):

PR #53, opened 2021-09-09
Kristo PDF TOC suggestions

kristojorg:
I just figured we could simplify this a little by putting the logic to fetch the current resource in one `useEffect` hook that responds to the change in resource index.


hokei:
ahhhh this is nice

PR #54, opened 2021-09-10
NO-REF Added direct link to fix parcel build errors

hokei:
running `parcel build` was throwing an error where it was looking inside the wrong root directory inside `parcel-reporter-static-files-copy`.  
This was stopping the build and making it so that the build was failing.  

My current theory is that it has something to do with this bug:  https://github.com/parcel-bundler/parcel/issues/6729 
However, making direct reference to the files seems to fix the problem.  

Would love insight into whether anyone else was having this error - it started appearing very recently for me, and this bug is about a month old.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "web-reader",
       "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
--- a/package.json
+++ b/package.json
@@ -8,10 +8,10 @@
     "src"
   ],
   "scripts": {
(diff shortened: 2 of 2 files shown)


kristojorg:
Ah shoot actually this isn't going to work. Running `parcel example/index.html` is a shortcut to run parcel on that one file. Running `parcel` plain however, looks in the `package.json` for a configuration. You can find our config under `targets` and `example-app` in `package.json`. I just undid your changes and am not getting the error you're getting so let's try to debug today

hokei:
Looks like this isn't going to work.  Closing

PR #55, opened 2021-09-13
Replace Axisnow example

kristojorg:
This PR replaces the Axisnow example publication with one that is appropriately open-access. In order to view it, you will have to now set `process.env.AXISNOW_ISBN` and `process.env.AXISNOW_VAULT_UUID` when starting the app. You can ask me for those values


hokei:
> You can ask me for those values

Any reason to make these values private?  Can the `AXISNOW_VAULT_UUID` be a fake one?  ISBN can definitely be public, right?

kristojorg:
> Any reason to make these values private? Can the AXISNOW_VAULT_UUID be a fake one? ISBN can definitely be public, right?

Leonard asked me to make the vault UUID private. I can put in a fake one but then the decryption wouldn't work. If we want it to work locally we will need to go in and change the value but remember not to check it in... so I figured easier just to make them env variables. I guess I could have left the ISBN public, but idk, just seemed to make sense to define both at the same time and place since you get them both when you check out the book

hokei (approved):
OK looks good.  I'd love to figure out if we could put the values in Confluence or a doc or something, so we're not relying on key telephone, but I can't figure out an obvious place to put it

PR #58, opened 2021-09-20
Update Build System

kristojorg:
This PR replaces our tsdx build system with a new one. It is sad that we have to do this because TSDX was the official typescript build recommendation, but it's been somewhat abandoned. I used [esbuild](https://esbuild.github.io) in R2D2BC and it worked really well. For one thing, it is way way wayyy faster than tsdx was, which actually makes a big productivity difference. 

There is another major difference though. It does not try to do everything. It just transpiles and bundles our JS/TS/CSS. Thus, it's not going to bloat like webpack into a thing with 1 billion options that make no sense. The way to use it is by writing a js or ts `build.ts` file that performs your build. Thus you have a ton of flexibility. You can do whatever else you need in that same script, whether it's copying static files or generating TS declaration files or whatever. Our build file is pretty simple and should hopefully be easier to read through and understand than a webpack or rollup config file because it's just typescript code. 

Anyway, I'm not necessarily suggesting this is the ultimate solution for us, just that we needed a new build setup to get this working for integration, and we can bikeshed and decide something different in the future if we need to.

For now, this is available on NPM via `@nypl/web-reader@0.2.0-alpha.0`. Give it a try in DRB @hokei and lmk how it goes

**CSS Setup**: I now bundle CSS for the html reader and pdf reader separately and you can import them like this: 
```ts
import "@nypl/web-reader/dist/PdfReader/styles.css";
import "@nypl/web-reader/dist/HtmlReader/styles.css"
```


build.ts
```diff
@@ -0,0 +1,172 @@
...
+import origRimraf from 'rimraf';
+import child_process from 'child_process';
+
+const rimraf = util.promisify(origRimraf);
+const exec = util.promisify(child_process.exec);
+
+/**
+ * @todo
+ *
+ * - Figure out how to export the fonts and include them with dynamic urls in the css?
```
kristojorg, line 16: These fonts were unused by our web reader but included in the ReadiumCSS. Not sure how they were dealt with before, possibly just uncaught errors that we didn't know about. Anyways, shouldn't be a problem for now

example/index.tsx
```diff
@@ -224,4 +224,53 @@ const AxisNowEncrypted: React.FC = () => {
   );
 };
 
+const url =
+  'http://localhost:3000/api/axisnow/9781429930543/879ED23D-B352-48CD-B4B9-65AE53CC6A20';
+const RemoteAxisNowEncrypted: React.FC = () => {
+  const [getContent, setGetContent] = React.useState<null | GetContent>(null);
+  const [error, setError] = React.useState<Error | undefined>(undefined);
+
+  const book_vault_uuid = '879ED23D-B352-48CD-B4B9-65AE53CC6A20';
```
hokei: is this ok to commit?

src/HtmlReader/ReadiumCss/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
```
hokei, line 1: I'm really confused - should readiumcss files live here, or should they live in the consuming app, or both? or is this a wip?

kristojorg: Sooo basically they're going to live in both for now. If the consuming app has a way to statically serve files from `node_modules` then they can use `import "@nypl/web-reader/dist/html-styles.css` and serve that statically. Next.js doesn't have a way to do that though, so I had to copy them over there for now.

src/HtmlReader/index.tsx
```diff
@@ -100,6 +100,13 @@ export default function useHtmlReader(args: ReaderArguments): ReaderReturn {
         navHeight: HEADER_HEIGHT,
         margin: 0,
       },
+      rights: {
+        /**
+         * Makes the reader fetch every resource before rendering, which
+         * takes forever.
+         */
+        autoGeneratePositions: false,
```
hokei, line 108: Are you setting this to false so that the reader doesn't do this, or is the reader currently doing this for some other reason?

kristojorg: This setting got defaulted to `true` in a recent update to R2D2BC. I don't think it should be that way because it makes the reader fetch every resource before rendering. I am here manually turning it off. We don't need it anyway for the features we are using.

hokei:
We've been using this build version for a while on our webreader integration and as far as I'm concerned, that means it's ok by me.  Why is the build step failing though?

kristojorg:
Ok I've fixed the CI error (missing dep) and re-enabled tests (though they are flaky and failing). The tests I am fixing in a separate branch so for now we will just have to merge this with them failing :/

samanthaandrews:
> Ok I've fixed the CI error (missing dep) and re-enabled tests (though they are flaky and failing). The tests I am fixing in a separate branch so for now we will just have to merge this with them failing :/

I think it's ok to merge with failing tests since you are fixing in a separate branch. Something that may help:

Adding retries to the `cypress.json` file:

```
  "retries": {
    "runMode": 2,
    "openMode": 0
  }
  
  ```
  This is what retries look like in CPW currently

hokei (approved):

PR #59, opened 2021-09-20
Removed font from PDF settings

Toxiapo:
This PR removes the font from the PDF settings module and updated some of the methods that were being referenced in the PDF reader.

- Removed font buttons from PDF settings card
- Renamed PDF methods. increaseFontSize, decreaseFontSize => zoomIn, zoomOut
- Updated SettingsCard test


src/PdfReader/index.tsx
```diff
@@ -400,8 +399,8 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
       navigator: {
         goForward,
         goBackward,
-        increaseFontSize,
-        decreaseFontSize,
+        zoomIn,
+        zoomOut,
         setFontFamily,
```
hokei: let's remove setFontFamily as well, since the pdf navigator doesn't need it.

hokei: (does the pdf navigator use setColorMode at all?)

src/ui/SettingsButton.tsx
```diff
@@ -25,8 +25,7 @@ export default function SettingsCard(
   const open = () => setIsOpen(!isOpen);
   const close = () => setIsOpen(false);
 
-  const { navigator, state, type } = props;
```
hokei, line 28: why get rid of this?

Toxiapo: Typescript was throwing errors about the Type after I separate the PdfNavigator and HtmlNavigator. 

I would get `Type 'PdfNavigator | HtmlNavigator' is not assignable to type 'PdfNavigator'.` error if I destructuring the props before passing it to the PdfSettings or HtmlSettings component. I think the reason was that at the time of destructuring, the `navigator` still has the type 'PdfNavigator | HtmlNavigator' but the settings component requires a specific navigator.

kristojorg: Exactly, when destructuring you lose some context typescript might have about the outer value (props in this case). Thus it becomes more difficult to refine these into one of the two types we need. It's easier in the end to keep it all together in `props`

hokei (approved):

src/PdfReader/index.tsx
hokei: this can also be moved to something else, it's not important.

Toxiapo: Removed setFontFamily. I don't see theme change being part of the PDF TAD so I removed setColorMode as well. 

The other ones can be addressed on [SFR-1288](https://jira.nypl.org/browse/SFR-1288) I think (e.g state values)

PR #62, opened 2021-09-24
SFR-1325  webreader SFR fixes

hokei:
Changes around the margins to integrate the PDF reader.


src/ui/manager.tsx
```diff
@@ -47,7 +47,6 @@ const WebReaderContent: React.FC<ReaderReturn & ReaderManagerArguments> = ({
         />
       </PageButton>
       <Flex
-        width="100vw"
```
hokei, line 50: This was removed because of a CSS issue that showed up in SFR.  I'm surprised that it didn't show up anywhere else - I did a cursory check to make sure it wasn't breaking anything in the example app, but please double check

Toxiapo: Yeah, it seems fine without the extra CSS.

kristojorg (approved):
Looks good to me, I haven't tested the CSS changes in CPW, but let's give it a publish and try it and if it doesn't work fix it

PR #64, opened 2021-09-25
WIP: (for discussion) allow overriding injectables

AntonOfTheWoods:
Currently you *must* (*) inject ReadiumCSS-after.css, ReadiumCSS-before.css, and
ReadiumCSS-default.css and they *must* be available at
your.site/css/XXX.css. The same is true of opendyslexic.css at
your.site/fonts/opendyslexic. It doesn't appear that any other files can
be injected, meaning it is impossible to extend the underlying reader
functionality.

This PR allows for the overriding of these values, and also injecting
`injectablesFixed`. This is the "cheap" option, though it means that
there are now quite a few type-specific props that are included in the
common `UseWebReaderArguments`, which is probably not ideal.

Feedback is requested before embarking on tests, etc.

(*) it appears it actually will "work" if rubbish is returned, as long
as the server responds with a 200. If the server responds with a 404, it
appears to fail

### DIFF
--- a/src/HtmlReader/index.tsx
+++ b/src/HtmlReader/index.tsx
@@ -1,6 +1,6 @@
 import D2Reader from '@d-i-t-a/reader';
 import React from 'react';
-import injectables from './injectables';
+import defaultInjectables from './injectables';
 import {
   ColorMode,
   HtmlReaderState,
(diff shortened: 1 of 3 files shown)


hokei:
Hi!  Sorry I merged the branch this was depending on, but IMO this is a super good idea, we'd wanted to do something about it, we really do want to get it in!  If you can, can you re-make it against `main` ?  

CC @kristojorg

AntonOfTheWoods:
Thanks @hokei https://github.com/NYPL-Simplified/web-reader/pull/69

PR #65, opened 2021-09-27
SFR-1322 Fixing cypress tests

samanthaandrews:
- Cleaning up Cypress tests to pass in CI

Jira issue: https://jira.nypl.org/browse/SFR-1322

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,3 +1,4 @@
 {
-  "baseUrl": "http://localhost:1234"
+  "baseUrl": "http://localhost:1234",
+  "projectId": "79fvf2"
 }
--- a/cypress/integration/html-reader/navigations.ts
+++ b/cypress/integration/html-reader/navigations.ts
@@ -9,18 +9,18 @@ describe('navigating an EPUB page', () => {
     cy.findByRole('link', { name: 'Return to Homepage' }).should(
       'have.prop',
       'href',
-      '/'
+      `${Cypress.config().baseUrl}/`
     );
   });
 
   it('should update page content after clicking on TOC link', () => {
-    cy.iframe(IFRAME_SELECTOR)
+    cy.getIframeBody(IFRAME_SELECTOR)
       .findByRole('img', {
         name: "Alice's Adventures in Wonderland, by Lewis Carroll",
       })
       .should('exist');
 
-    cy.iframe(IFRAME_SELECTOR)
+    cy.getIframeBody(IFRAME_SELECTOR)
       .findByText('Down the Rab­bit-Hole')
       .should('not.exist');
 
@@ -34,14 +34,14 @@ describe('navigating an EPUB page', () => {
     cy.get('#reader-loading').should('be.visible');
     cy.get('#reader-loading').should('not.be.visible');
 
-    cy.iframe(IFRAME_SELECTOR)
+    cy.getIframeBody(IFRAME_SELECTOR)
       .findByText('Down the Rab­bit-Hole')
       .should('exist');
   });
 
   it('should navigate forward and backwards with page buttons', () => {
     cy.log('make sure we are on the homepage');
-    cy.iframe(IFRAME_SELECTOR)
+    cy.getIframeBody(IFRAME_SELECTOR)
       .findByRole('img', {
         name: "Alice's Adventures in Wonderland, by Lewis Carroll",
       })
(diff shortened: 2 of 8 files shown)


cypress/integration/html-reader/navigations.ts
```diff
@@ -9,18 +9,18 @@ describe('navigating an EPUB page', () => {
     cy.findByRole('link', { name: 'Return to Homepage' }).should(
       'have.prop',
       'href',
-      '/'
+      `${Cypress.config().baseUrl}/`
```
samanthaandrews, line 12: This href is expecting the full path, so I added the base url since that's what it is expecting

cypress/integration/html-reader/render-app.ts
```diff
@@ -1,22 +1,24 @@
...
-//   });
-// });
+    // Page one contains an image
+    cy.getIframeBody(IFRAME_SELECTOR)
+      .find('img')
+      .should(
+        'have.attr',
+        'alt',
+        "Alice's Adventures in Wonderland, by Lewis Carroll"
+      );
```
samanthaandrews, line 22: Tbh I have no idea why this is less flaky than what was previously written. I had the same experience with the iframe tests in CPW.

cypress/support/commands.ts
```diff
@@ -24,17 +21,18 @@ Cypress.Commands.add('loadPage', (pageName) => {
...
+    .should('not.be.empty')
+    .then(cy.wrap);
+});
 
+Cypress.Commands.add('getIframeHtml', (selector: string) => {
   return cy
-    .get(selector, { timeout: ops.timeout })
-    .its(`0.contentDocument`)
+    .get(selector, { timeout: 15000 })
+    .its('0.contentDocument.documentElement')
```
samanthaandrews: The reason some of these tests were flaky is bc we were looking for the `html` element within the iframe, but that element isn't natively available within an iframe. Instead we can use `documentElement`:

> Document.documentElement returns the Element that is the root element of the document (for example, the <html> element for HTML documents).

https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement

cypress/support/commands.ts
```diff
@@ -24,17 +21,18 @@ Cypress.Commands.add('loadPage', (pageName) => {
...
-  };
-
-  const ops = { ...DEFAULT_OPTIONS, ...options };
+Cypress.Commands.add('getIframeBody', (selector: string) => {
+  return cy
+    .get(selector, { timeout: 15000 })
+    .its(`0.contentDocument.body`)
+    .should('not.be.empty')
+    .then(cy.wrap);
+});
```
samanthaandrews: I think being explicit about which part of the iframe we are selecting is better for these commands. Makes things more clear.

cypress/integration/html-reader/navigations.ts
```diff
@@ -9,80 +9,91 @@ describe('navigating an EPUB page', () => {
...
-    cy.get('#reader-loading').should('be.visible');
-    cy.get('#reader-loading').should('not.be.visible');
+    cy.wait('@chapterOne', { timeout: 10000 }).then((interception) => {
+      assert.isNotNull(
+        interception?.response?.body,
+        'chapter one API call has data'
+      );
+    });
+
+    cy.wait(3000);
```
samanthaandrews, line 42: Even after adding the intercepts, I still have to add these wait() commands. This is because the following command is running before the iframe has fully updated, so it is flaky. In CPW I mitigated this by checking the iframe src attribute, but that same attribute doesn't exist in the new web reader. So I think wait() is our best bet for now.

cypress.json
```diff
@@ -1,3 +1,4 @@
 {
-  "baseUrl": "http://localhost:1234"
+  "baseUrl": "http://localhost:1234",
+  "projectId": "79fvf2"
```
samanthaandrews, line 3: I created a Cypress Dashboard project for this repo :D

samanthaandrews:
Ok these tests are passing now.

In order to finish setting up the cypress dashboard, however, I'll need access to this repo's GH secrets so I can add the record key.

cypress/support/commands.ts
```diff
@@ -24,17 +22,26 @@ Cypress.Commands.add('loadPage', (pageName) => {
...
 
-Cypress.Commands.add('iframe', (selector: string, options?: IframeOptions) => {
-  const DEFAULT_OPTIONS = {
-    // Cypress default timeout
-    timeout: 15000,
-  };
+Cypress.Commands.add('getIframeHtml', (selector: string) => {
+  return cy
+    .get(selector, { timeout: 15000 })
+    .its('0.contentDocument.documentElement')
```
samanthaandrews, line 28: https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement

Toxiapo (approved):
LGTM! Thanks for cleaning this up!

PR #66, opened 2021-09-27
Fix TOC Rendering

Toxiapo:
This PR fixes the TOC rendering with titles containing HTML. The rendered text should be human-readable text instead of Html entities (e.g `&#8212;`)

### DIFF
--- a/src/ui/TableOfContent.tsx
+++ b/src/ui/TableOfContent.tsx
@@ -15,55 +15,6 @@ import useColorModeValue from './hooks/useColorModeValue';
 import { ReadiumLink } from '../WebpubManifestTypes/ReadiumLink';
 import { HEADER_HEIGHT } from './constants';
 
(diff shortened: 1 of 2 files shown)


src/ui/TableOfContent.tsx
```diff
@@ -142,3 +93,56 @@ export default function TableOfContent({
...
+  return (
+    <MenuItem as={Link} href={href} {...styles} {...rest}>
+      <RenderHtml title={title ?? ''} />
+    </MenuItem>
+  );
+};
+
+const RenderHtml = ({ title }: { title: string }) => (
+  <span dangerouslySetInnerHTML={{ __html: title }}></span>
+);
```
Toxiapo, line 148: I had to create a separate component for the HTML component because I was getting [Can only set one of `children` or `props.dangerouslySetInnerHTML`](https://stackoverflow.com/a/39910404/8606992) error.

kristojorg (approved):
Looks good to me!

PR #68, opened 2021-09-29
SFR-1290 single pdf toc

hokei:
This PR injects the PDFjs Outline to the Table of Contents.  
Wanted to get some feedback on the organization of `TableOfContent.tsx` - it's now starting to look pretty messy, and I'm wondering if it's going to be easier to have a separate `PDFOutlineToc` and a `WebpubToc` but have it share the same styling.

Still outstanding: 
- [ ] The current chapter should be highlighted on PDFs.  (New PR - the solution to this might be tied to the "next chapter button" issue.  We will need to figure out a way to map page numbers to dest items, something that currently doesn't exist. 
- [ ] Tests for TOC functionality, in Epub, Multi PDF and Single PDF instances  -- pushed onto @samanthaandrews 
- [x] Sub-chapters don't link correctly 
- [x] There's a bunch of weird merge things going on here that I should fix


.storybook/preview.jsx
```diff
@@ -1,23 +1,21 @@
-const React = require('react');
-const { ChakraProvider } = require('@chakra-ui/react');
-import { getTheme } from '../src/ui/theme';
-import '@nypl/design-system-react-components/dist/styles.css';
+const React = require("react")
```
hokei: Looks like there's stuff my linter is doing differently ... what settings are y'all using for jsx?

package.json
```diff
@@ -24,7 +24,7 @@
   "scripts": {
     "example:build": "parcel build",
     "example:nocache": "parcel --no-cache",
-    "example": "parcel",
+    "example": "parcel example/index.html",
```
kristojorg, line 27: Hopefully we don't still need this? It overrides the settings at line 68 in this file

samanthaandrews:
> Wanted to get some feedback on the organization of TableOfContent.tsx - it's now starting to look pretty messy, and I'm wondering if it's going to be easier to have a separate PDFOutlineToc and a WebpubToc but have it share the same styling.

This is looking really good! I do think splitting up PDFOutlineToc and WebpubToc makes sense for clarity.

src/types.ts
```diff
@@ -38,8 +39,22 @@ export type ReaderState = {
...
+  url: string | null;
+  unsafeUrl: string | undefined;
+  newWindow: boolean | undefined;
+  count: number | undefined;
+  items: PDFTreeNode[];
+};
+
 // PDF specific reader state
-export type PdfReaderState = ReaderState;
+export type PdfReaderState = ReaderState & { pdf?: PDFDocumentProxy };
```
kristojorg: If I'm not wrong, this should be added inside `/PdfReader/index.tsx` in the `PdfState`. The `PdfReaderState` type here is meant to be the public-facing state that is returned by `usePdfReader`, but I don't think that consumers of `usePdfReader` would want access to that? The `PdfState` is where I keep state that is internal to the component. Confusing, I know. I should do rename these things

src/PdfReader/index.tsx
```diff
@@ -329,25 +331,53 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
   }, [state.scale]);
 
+  // If this is a single PDF webpub, goToPage navigates to the page using the destination array.
+  // If this is a multi PDF webpub, goToPage loads the new resource.
   const goToPage = React.useCallback(
-    async (href) => {
-      const getIndexFromHref = (href: string): number => {
-        const index = manifest?.readingOrder?.findIndex((link) => {
-          return link.href === href;
+    (target: string | any[]) => {
```
kristojorg: This works for now, but I think in the future we should try to make this work with a Readium `Locator` object, which I believe is supposed to handle cases where it is a page or specific location within a resource. Thus we can avoid having the react-pdf specific logic and types leak out of `usePdfReader`

kristojorg: [Here is locator reference](https://readium.org/architecture/models/locators/)

kristojorg: This is what a pdf locator would look like:
```
{
  "href": "http://example.com/document",
  "type": "application/pdf",
  "title": "Page 5",
  "locations": {
    "fragments": ["page=5", "viewrect=50,50,640,480"],
    "progression": 0.12703,
    "totalProgression": 0.12703
  }
}
```

src/ui/TableOfContent.tsx
```diff
@@ -9,33 +9,151 @@ import {
...
+    evt: React.MouseEvent<Element, MouseEvent>,
+    item?: PDFTreeNode
+  ) => {
     evt.preventDefault();
-    const href = evt.currentTarget.getAttribute('href');
-    if (!href) {
-      console.warn('TOC Link clicked without an href');
-      return;
+    const href: string | null = evt.currentTarget.getAttribute('href');
+    console.log('href', href);
```
kristojorg: stray console logs

src/ui/TableOfContent.tsx
```diff
@@ -9,33 +9,151 @@ import {
...
+      } else {
+        throw new Error('trying to get state');
+      }
+    };
+
+    if (outline.length === 0) {
+      getOutline();
+    }
+  }
+
```
kristojorg: Question: 

Is it possible in `usePdfReader` to process the pdf outline into a readium-style `toc`? Then we could just add it to the webpub manifest we return and render it normally using this component

kristojorg (changes requested):
I think we should try to keep the pdf-specific toc stuff out of our ui components by using the Readium `Locator` abstraction or even just url hashes: `goTo('resourceUrl#page=2')`. Let's huddle about this today if we can!

kristojorg:
Also yea idk what's going on with the linter, it should be using the settings found in the repo, and also should lint fix before committing

package.json
hokei: Oh right thanks for pointing that out - I am still getting errors when I run locally so I just put it in when developing.

src/ui/TableOfContent.tsx
```diff
@@ -10,33 +10,150 @@ import {
...
-    const href = evt.currentTarget.getAttribute('href');
-    if (!href) {
-      console.warn('TOC Link clicked without an href');
-      return;
+    const href: string | null = evt.currentTarget.getAttribute('href');
+    if (href && href !== '#') {
+      navigator.goToPage(href);
+      setIsOpen(false);
+    } else if (item && item.dest) {
+      console.log('item', item);
```
samanthaandrews, line 97: stray `console.log()`

samanthaandrews:
https://user-images.githubusercontent.com/34348332/136277946-1f2e635a-ee9e-4e09-8a5c-d8aed8297128.mov

I'm not sure what the expected behavior is in scroll mode, but I attached a recording of what I noticed when switching from paginated to scrolling mode and then attempting to use the TOC to navigate to a section.

hokei:
@samanthaandrews oh no that's defs not good.  Thanks for finding it! 
I feel a bit sheepish for never having tested it in scroll mode - but I think the toc doesn't work in scroll mode at all due to the fact that both this TOC functionality, and also scroll mode, are both kind of hacky.  fun.

hokei:
Closing this PR pending a re-thinking of this work - I will comment this on JIRA ticket so we can refer in the future.

PR #69, opened 2021-09-30
WIP: (for discussion) allow overriding injectables

AntonOfTheWoods:
Currently you *must* (*) inject ReadiumCSS-after.css, ReadiumCSS-before.css, and
ReadiumCSS-default.css and they *must* be available at
your.site/css/XXX.css. The same is true of opendyslexic.css at
your.site/fonts/opendyslexic. It doesn't appear that any other files can
be injected, meaning it is impossible to extend the underlying reader
functionality.

This PR allows for the overriding of these values, and also injecting
`injectablesFixed`. This is the "cheap" option, though it means that
there are now quite a few type-specific props that are included in the
common `UseWebReaderArguments`, which is probably not ideal.

Feedback is requested before embarking on tests, etc.

(*) it appears it actually will "work" if rubbish is returned, as long
as the server responds with a 200. If the server responds with a 404, it
appears to fail

### DIFF
--- a/src/HtmlReader/index.tsx
+++ b/src/HtmlReader/index.tsx
@@ -1,6 +1,6 @@
 import D2Reader from '@d-i-t-a/reader';
 import React from 'react';
-import injectables from './injectables';
+import defaultInjectables from './injectables';
 import {
   ColorMode,
   HtmlReaderState,
@@ -74,7 +74,13 @@ function htmlReducer(state: HtmlState, action: HtmlAction): HtmlState {
 const FONT_SIZE_STEP = 4;
 
 export default function useHtmlReader(args: ReaderArguments): ReaderReturn {
-  const { webpubManifestUrl, manifest, getContent } = args ?? {};
+  const {
+    webpubManifestUrl,
+    manifest,
+    injectables,
+    injectablesFixed,
+    getContent,
+  } = args ?? {};
   const [state, dispatch] = React.useReducer(htmlReducer, {
     colorMode: 'day',
     isScrolling: false,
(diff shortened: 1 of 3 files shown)


kristojorg:
This is great, the one thing I want to try and figure out is how we separate web reader arguments that _only_ apply to the HTML reader, such as injectables in this case. They aren't injected in the PDF reader. Ideally it wouldn't be possible (from a type perspectivE) to supply invalid arguments, but I'll have to think about how that can be done. 

One thing I am considering is that maybe users always know at runtime whether they are rendering a PDF Webpub or an HTML Webpub, and thus could import and use two separate components: `<HtmlReader>` or `<PdfReader>`. Then, they would just have different sets of accepted props. It just depends if we want to continue allowing the generic `<WebReader>` that applied to both formats...

AntonOfTheWoods:
> This is great, the one thing I want to try and figure out is how we separate web reader arguments that _only_ apply to the HTML reader, such as injectables in this case. They aren't injected in the PDF reader. Ideally it wouldn't be possible (from a type perspectivE) to supply invalid arguments, but I'll have to think about how that can be done.

Thanks, this was my thought also, and why I wanted your feedback before doing anything more substantial (i.e, adding a test or two). The only things I can think of aren't particularly elegant, but if I think of something better I'll suggest it!

kristojorg:
Ok I ended up incorporating these suggestions into #72 , and have published `@nypl/web-reader@0.2.0-alpha.11` with these changes. 

Thanks for your contribution here @antonofthewoods!

PR #70, opened 2021-09-30
Fix ServiceWorker Cacheing

kristojorg:
In CPW when using the `usePublicationSW` hook, I was experiencing delays which I believe were caused by the UI thread blocking while the hook requested all these resources and cached them. Therefore, I have moved the actual requesting and cacheing into the service worker itself. We now just send it a message telling it which publications to cache. 

I also bail out now if the resource is already in the cache. It will be updated when the cache entry expires.

Hopefully his will improve performance in CPW, but even if not it is a better practice anyway to do this work off the main thread.


samanthaandrews:
@kristojorg Let's huddle after standup on Monday (@hokei and @Toxiapo welcome to join too!) to review this and https://github.com/NYPL-Simplified/circulation-patron-web/pull/290 together. It's all making sense to me thus far, but want to make sure I understand how the pieces fit together.

samanthaandrews (approved):
LGTM

PR #71, opened 2021-10-01
Add aria-label to radio buttons

Toxiapo:
This PR adds aria-label to the radio buttons because we were not able to access these elements by their labels, mostly due to the fact that we have a nested `div` inside the label element, making the accessibility name to be null when using APIs like `getByRole`.

- Added aria-label to the ToggleButton component
- Updated Settings card tests


samanthaandrews (approved):
yayy this looks great thanks @Toxiapo

PR #72, opened 2021-10-02
Fix loading indicator

kristojorg:
This PR fixes the loading indicator by importing the R2D2BC css into the html reader. This gets compiled to `/dist/esm/index.css`, so consumers will need to import it from there. 

I have also figured out how to bundle the readium css files into `dist/injectable-html-styles.css`, which can then be imported in a consuming app as a url and passed to the web reader as an injectable. To make this happen, I have also implemented the suggested injectable customization from #69. 

This injectable customization will work for now, but I can see a future where the options for PDFs and HTML publications diverge. It may make sense to simply split them into two exported components using the same UI components and architecture.


example/index.tsx
```diff
@@ -25,11 +25,25 @@ import {
 import { getTheme } from '../src/ui/theme';
 import usePublicationSW from '../src/ServiceWorker/client';
 import AxisNowEncrypted from './axisnow-encrypted';
+import htmlStyles from 'url:../src/HtmlReader/injectable-styles.css';
```
kristojorg, line 27: This is one way to import a css file as a url which works in parcel. In webpack you would use `file-loader`

samanthaandrews (approved):
LGTM

samanthaandrews:
I did notice that PDF loading indicator still looks like this. I believe this is intended, but wanted to make sure.

Screen Shot 2021-10-03 at 6 08 18 PM

example/index.tsx
```diff
@@ -212,7 +231,9 @@ const App = () => {
 const DynamicReader: React.FC = () => {
   const { manifestUrl } = useParams<{ manifestUrl: string }>();
   const decoded = decodeURIComponent(manifestUrl);
-  return <WebReader webpubManifestUrl={decoded} />;
+  return (
+    <WebReader injectables={htmlInjectables} webpubManifestUrl={decoded} />
```
AntonOfTheWoods, line 239: @kristojorg Thanks for this! I'm not totally clear on how these will get passed to the `useHtmlReader` though, as I can't see the `...props` getting passed. Am I missing something, or should these be somehow passed to https://github.com/NYPL-Simplified/web-reader/blob/main/src/useWebReader.tsx#L55?

kristojorg: Nope you are entirely correct, I was seeing it working in my npm linked project due to browser cacheing. Thanks for catching this! I should really write some tests for this functionality.

kristojorg:
Ok I added some tests to make sure the injectables were properly passed on to the reader after @AntonOfTheWoods caught a mistake I made : ). 

@samanthaandrews you're right the PDF reader still has that loader, but that's going to be a separate ticket.

I'll merge this once tests pass!

samanthaandrews:
I refactored these tests a little and moved them into their own spec. For some reason, moving to their own spec seemed to solve the issue. Separating specs when things are behaving unexpectedly in CI is something recommended by Cypress docs.

With the `getIframeHead()` and other `getIframe` custom commands, we don't need to use "within" because we return the wrapped element, which means we can chain directly off that command.

cypress/integration/html-reader/injectables.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  it('should render css injectables when provided', () => {
+    cy.visit('/test/with-injectables');
+    cy.getIframeHead(IFRAME_SELECTOR).contains('title', 'Cover');
+    cy.getIframeHead(IFRAME_SELECTOR)
+      .find(
+        `link[href="${
+          Cypress.config().baseUrl
+        }/fonts/opensyslexic/opendyslexic.css"]`
+      )
```
kristojorg, line 18: So was this the key here @samanthaandrews? Using `.find` instead of `.within` and `.get`?

kristojorg: Also I previously had `href^=...` in order to match an href containing the string so I didn't have to provide the baseUrl and everything. Did that not work for you?

PR #75, opened 2021-10-07
SFR-1347: Adding PDF Cypress tests

samanthaandrews:
- Adds Cypress tests for single-pdf and multi-pdf.
- Ensures that these resources are loading as expected, that you can navigate with pagination, scrolling, and TOC, and that toggling settings makes the appropriate updates.

### DIFF
--- a/cypress/integration/html-reader/injectables.ts
+++ b/cypress/integration/html-reader/injectables.ts
@@ -11,14 +11,10 @@ describe('useHtmlReader configuration settings', () => {
     cy.visit('/test/with-injectables');
     cy.getIframeHead(IFRAME_SELECTOR).contains('title', 'Cover');
     cy.getIframeHead(IFRAME_SELECTOR)
-      .find(
-        `link[href="${
-          Cypress.config().baseUrl
-        }/fonts/opensyslexic/opendyslexic.css"]`
-      )
+      .find(`link[href$="/fonts/opensyslexic/opendyslexic.css"]`)
       .should('exist');
     cy.getIframeHead(IFRAME_SELECTOR)
-      .find(`link[href="${Cypress.config().baseUrl}/css/sample.css"]`)
+      .find(`link[href$="/css/sample.css"]`)
       .should('exist');
     cy.getIframeBody(IFRAME_SELECTOR).should(
       'have.css',
--- a/cypress/integration/single-pdf/render-app.ts
+++ b/cypress/integration/single-pdf/render-app.ts
@@ -1,0 +1,15 @@
+describe('Renders single PDF', () => {
+  it('Render page content', () => {
+    cy.loadPdf('/pdf');
+
+    cy.log('check that all the essential buttons are on the page');
+    cy.findByRole('link', { name: 'Return to Homepage' }).should('exist');
+    cy.findByRole('button', { name: 'Table of Contents' }).should('exist');
+    cy.findByRole('button', { name: 'Settings' }).should('exist');
+    cy.findByRole('button', { name: 'Toggle Fullscreen' }).should('exist');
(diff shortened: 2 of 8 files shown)


cypress/integration/html-reader/injectables.ts
```diff
@@ -11,14 +11,10 @@ describe('useHtmlReader configuration settings', () => {
...
-      .find(
-        `link[href="${
-          Cypress.config().baseUrl
-        }/fonts/opensyslexic/opendyslexic.css"]`
-      )
+      .find(`link[href$="/fonts/opensyslexic/opendyslexic.css"]`)
       .should('exist');
     cy.getIframeHead(IFRAME_SELECTOR)
-      .find(`link[href="${Cypress.config().baseUrl}/css/sample.css"]`)
+      .find(`link[href$="/css/sample.css"]`)
```
samanthaandrews, line 17: @kristojorg changed these back to your original code. this was never the problem, i just got annoyed and started changing lots of things to get it to work!

cypress/integration/single-pdf/render-app.ts
```diff
@@ -0,0 +1,15 @@
...
+
+    cy.log('check that all the essential buttons are on the page');
+    cy.findByRole('link', { name: 'Return to Homepage' }).should('exist');
+    cy.findByRole('button', { name: 'Table of Contents' }).should('exist');
+    cy.findByRole('button', { name: 'Settings' }).should('exist');
+    cy.findByRole('button', { name: 'Toggle Fullscreen' }).should('exist');
+    cy.findByRole('button', { name: 'Next Page' }).should('exist');
+    cy.findByRole('button', { name: 'Previous Page' }).should('exist');
+
+    cy.findByText('Julius H. Schoeps').should('be.visible');
```
hokei, line 13: I'm surprised that this doesn't time out because the PDF takes so long to load -- would it make sense to switch this out for a smaller pdf if we find one?

cypress/integration/single-pdf/settings.ts
```diff
@@ -0,0 +1,43 @@
...
+      'not.exist'
+    );
+    cy.findByRole('button', { name: 'Decrease font size' }).should('not.exist');
+    cy.findByRole('button', { name: 'Increase font size' }).should('not.exist');
+  });
+
+  it('should zoom in and out', () => {
+    cy.findByText('Julius H. Schoeps')
+      .should('be.visible')
+      .should('have.css', 'font-size', '8.9176px');
```
hokei, line 27: are we worried at all about this being brittle?

cypress/integration/single-pdf/render-app.ts
samanthaandrews: I wrote a custom command to load the pdf `loadPdf()` above, which is coming from the `commands.ts` file. In this function I am making sure the GET request for the pdf resource returns before moving forward with the test. That being said, I did set a 20 second timeout for that PDF resource. It usually completes faster than 20 seconds, but figured it was good to give it some wiggle room. Smaller PDF would likely make these tests run faster.

cypress/integration/single-pdf/settings.ts
samanthaandrews: I'm not sure. We are re-setting the session storage every time the test runs (in the `loadPdf()` function), so it wouldn't be brittle due to any other tests or caching issues. Could it be brittle based on something that changes within the pdf resource that we don't control?

hokei: I think if it's not causing issues we can leave it.  Usually when things are this precise it sets off "brittle!!!" flags in my head, but it also seems pretty easy to make more forgiving if we need to.

hokei (approved):
These look really good to me!

PR #79, opened 2021-10-14
Add overflow to the iframe-wrapper element 

Toxiapo:
This PR adds the `overflow: auto` to the iframe-wrapper element so that the page button works on firefox.


hokei:
this is going to need the new R2 to start working right?  I'm still seeing the bug in the preview deploy and we shouldn't merge until we can verify

hokei (approved):

PR #80, opened 2021-10-14
SFR-1348: Create getContent() Cypress tests

samanthaandrews:
1. Added cypress tests for `getContent()`
2. Changed existing cypress html-reader tests to use the Moby Dick example
3. Upgrade Cypress to 8.6.0

https://jira.nypl.org/browse/SFR-1348


cypress/support/commands.ts
```diff
@@ -17,6 +17,12 @@ declare global {
 }
 
 Cypress.Commands.add('loadPage', (pageName) => {
+  cy.intercept('/samples/**/*', { middleware: true }, (req) => {
+    req.on('before:response', (res) => {
+      // force all API responses to not be cached
+      res.headers['cache-control'] = 'no-store';
+    });
+  });
```
samanthaandrews: This ensures network requests aren't being cached in tests, which makes waiting for network intercepts more reliable. I added the same block to the `loadPdf()` function below.

cypress/integration/axisnow-encrypted-epub/navigations.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    cy.wait(3000);
+
+    cy.getIframeHead(IFRAME_SELECTOR).contains('title', 'The Mail');
+    cy.getIframeBody(IFRAME_SELECTOR)
+      .findByText('The Mail')
+      .should('be.visible');
+    cy.getIframeBody(IFRAME_SELECTOR)
+      .findByText('“Halloa!” the guard replied.')
+      .should('exist');
```
samanthaandrews: Testing scrolling versus pagination mode has been difficult. Because of all the important content is within the iframe, I can't seem to use Cypress' default `scrollTo()` api. It always just says that my content is not scrollable. If anyone has any ideas for how to test that scroll mode is working as expected, I'm open!

cypress/integration/axisnow-encrypted-epub/navigations.ts
```diff
@@ -0,0 +1,114 @@
...
+        'chapter API call has data'
+      );
+    });
+
+    cy.wait(3000);
+
+    cy.getIframeHead(IFRAME_SELECTOR).contains('title', 'The Mail');
+    cy.findByRole('button', { name: 'Next Page' }).click();
+    // FIXME: Turning pages is not working as expected in paginated mode.
+    // Update this test as we fix https://jira.nypl.org/browse/SFR-1364
```
samanthaandrews: Writing this test is where I noticed that pagination and displaying chapters correctly isn't working. @Toxiapo agreed do look into this.

samanthaandrews:
display settings -- should update the font family to serif, paginated mode, and on sepia theme (failed)

Does this mean that the vercel preview deploy can't access our private NPM package? (I commented out these tests for now until I can get these pages to load in the preview deploys) @hokei @kristojorg Do we need to start passing a github token to our Vercel deploys?

hokei:
> Does this mean that the vercel preview deploy can't access our private NPM package? 

The vercel preview deploy /should/ have access, but it doesn't currently, it seems.  I know it's possible because we're deploying to openebooks.net through Vercel, though, so I'll look into it. 

Edit:  I was mixing up this repo and the CPW repo - I'm wondering if the AxisNow tests should live in this repo or if they should live with the CPW tests?  On one hand, it would be good to know that non-AxisNow work isn't breaking AxisNow work, and to know that this all continues to function.  On the other hand, it's incredibly frustrating to expect people to fix test failures in the parts of the app that they don't have access to.  @kristojorg thoughts?

kristojorg:
Yea I was thinking about this and I think best case scenario is that this test suite does have one axisnow based test. An alternative is stubbing out the decryptor in the tests so we don't test actual decryption. But if we are testing against the preview that won't work.

I think in the end the complexity of setting up tests for axisnow here outweigh the benefits for the time being. Let's keep them in CPW and maybe add a simple test to ensure the `getContent` api works without actually decrypting anything. Maybe one where `getContent` just returns some simple html always.

samanthaandrews:
> I think in the end the complexity of setting up tests for axisnow here outweigh the benefits for the time being. Let's keep them in CPW and maybe add a simple test to ensure the `getContent` api works without actually decrypting anything. Maybe one where `getContent` just returns some simple html always.

@kristojorg I'm putting some time on your calendar for Wednesday to pair on this. I attempted a version of it but not sure I'm heading in the right direction.

cypress/integration/html-reader/navigations.ts
```diff
@@ -13,89 +13,50 @@ describe('navigating an EPUB page', () => {
...
 
-    cy.wait('@imprint', { timeout: 10000 }).then((interception) => {
-      assert.isNotNull(
-        interception?.response?.body,
-        'imprint API call has data'
-      );
-    });
-
-    cy.wait(3000);
+    // FIXME: Uncomment the following once https://jira.nypl.org/browse/SFR-1332 is resolved
```
samanthaandrews, line 33: Currently can't navigate using the "next page" button because of this known issue.

cypress/integration/html-reader/toc-navigation.ts
```diff
@@ -0,0 +1,85 @@
+import { IFRAME_SELECTOR } from '../../support/constants';
+
+describe('navigating an EPUB page', () => {
+  beforeEach(() => {
+    cy.loadPage('/streamed-alice-epub');
```
samanthaandrews, line 5: Ideally I would make all the tests use the `/moby-epub` example, but in this case I was having less flake for these specific tests using the alice in wonderland example, so I kept it as is.

cypress/integration/multi-pdf/navigations.ts
```diff
@@ -22,12 +23,11 @@ describe('Multi PDF navigation', () => {
...
     cy.findByRole('button', { name: 'Table of Contents' }).click();
     cy.findByRole('menuitem', { name: 'Foreword' }).click();
+    cy.wait(5000);
     cy.findByText('Anthropology without Informants', { timeout: 10000 }).should(
       'be.visible'
     );
-    cy.get('div[data-page-number="2"]').contains('Foreword', {
-      timeout: 10000,
-    });
+    cy.get('div[data-page-number="2"]').should('exist');
```
samanthaandrews, line 30: This check for the word 'Foreword' became flaky for some reason so I am now just checking that this div is present in the dom.

cypress/support/commands.ts
```diff
@@ -53,13 +76,18 @@ Cypress.Commands.add('loadPdf', (path: '/pdf' | '/pdf-collection') => {
     Cypress.config().baseUrl === 'http://localhost:1234'
       ? 'http://localhost:3001'
       : 'https://drb-api-qa.nypl.org/utils';
-  cy.intercept('GET', `${pdfProxyInterceptUrl}/**`).as('pdf');
+  cy.intercept(`${pdfProxyInterceptUrl}/**`, { middleware: true }, (req) => {
+    req.on('before:response', (res) => {
+      // force all API responses to not be cached
+      res.headers['cache-control'] = 'no-store';
+    });
+  }).as('pdf');
```
samanthaandrews, line 84: this ensures we aren't caching resources in tests, which can cause cypress to not know that fetch requests have happened successfully

cypress/integration/html-reader/toc-navigation.ts
kristojorg: Werid... why flake why

kristojorg (approved):
LGTM!

PR #81, opened 2021-10-15
Update r2d2bc

Toxiapo:
This RP updates R2D2BC v2 which addresses the next page button inoperable and the blank page on paginated mode bug.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
   },
   "dependencies": {
     "@chakra-ui/react": "^1.6.6",
-    "@d-i-t-a/reader": "^2.0.0-beta.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -143,7 +143,7 @@
   },
   "dependencies": {
     "@chakra-ui/react": "^1.6.6",
-    "@d-i-t-a/reader": "2.0.0-alpha.7",
+    "@d-i-t-a/reader": "^2.0.0-beta.2",
```
hokei: we should probably not caret this -  we don't want it to automatically pull in potentially unstable updates

hokei (approved):

PR #84, opened 2021-10-19
Add getTheme export

### LINKED ISSUE #77: Further examples using the npm package rather than local paths
I have been trying to use the `ChakraProvider` theme system but it seems to rely on the `getTheme` method which I am unable to consume. I apologise if this is obvious, I am still relatively novice with modern JS.

I have created an example repo using `create-react-app` for typescript, with a minimal example that I hoped would consume the https://alice.dita.digital/manifest.json manifest.

https://github.com/AntonOfTheWoods/web-reader-test

The error is : Module not found: Can't resolve '@nypl/web-reader/dist/types/ui/theme' in '.../my-app/src'

Am I missing something here? Are we not supposed to be able to use the theme in this way? Are we supposed to implement our own system somehow? Thanks!

AntonOfTheWoods:
This allows for easy reuse of the theming system

Fixes https://github.com/NYPL-Simplified/web-reader/issues/77

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -35,3 +35,4 @@ export { usePublicationSW } from './ServiceWorker/index';
 export { default as useWebReader } from './useWebReader';
 export { default as useHtmlReader } from './HtmlReader';
 export { default as usePdfReader } from './PdfReader';
+export { getTheme } from './ui/theme';


kristojorg (approved):
I like it 👍  thanks for the PR!

AntonOfTheWoods:
Thanks for the review @kristojorg . Is there something else someone needs to do or is this a flaky tests issue?

kristojorg:
@AntonOfTheWoods I'm investigating that today

kristojorg:
closed by https://github.com/NYPL-Simplified/web-reader/pull/90 . The tests not running were caused by the fact that I had to manually approve the preview deployment from your branch, but then it was too late for the tests to detect it I guess.

PR #85, opened 2021-10-19
SFR 1335 - Add error states

kristojorg:
This PR adds an error boundary to the default export, but also adds a secondary (named) export without the error boundary in case a consuming application would like to customize the error handling. I also add some better errors for certain states and add some tests for the error states. We should continue to add error states and tests for them as the need arises using this framework.


kristojorg:
Not sure what happened with my commits here... But the files changed look correct to me

README.md
```diff
@@ -113,6 +118,18 @@ Finally, to use in a vanilla Javascript app:
...
 ```
 
+## Styling
+
+We ship a css file with each entrypoint that should be imported. This is necessary
+
+## Injectables
+
+The HTML Reader has the ability to inject custom elements into the reader iframe. This is most useful for passing
+
```
Toxiapo: unfinished sentences?

src/ui/Header.tsx
```diff
@@ -43,21 +43,7 @@ export default function Header(
...
-      height={`${HEADER_HEIGHT}px`}
-      zIndex="sticky"
-      alignContent="space-between"
-      alignItems="center"
-      px={8}
-      borderBottom="1px solid"
-      borderColor="gray.100"
-      bgColor={mainBgColor}
-    >
+    <HeaderWrapper bgColor={mainBgColor}>
```
Toxiapo, line 48: `HeaderWrapper` component is missing this bgColor style.  I think this is needed or else you will get a weird transition bg color.

README.md
```diff
@@ -2,7 +2,7 @@
 
 This project is a web reader built by NYPL for reading eBooks. It is built using the [Readium Architecture](https://github.com/readium/architecture), and specifically built for Webpubs. Webpub is a spec [defined by the Readium Foundation](https://github.com/readium/webpub-manifest) to provide a common abstraction between many types of web publications. Initially, this project will focus on HTML-based Webpubs and Webpubs that define PDF collections. An HTML-based Webpub can be generated from many types of eBooks, but most commonly ePubs.
 
-The project is bootstrapped with [TSDX](https://tsdx.io). It uses Typescript, React, Jest and Rollup, and features both a Storybook development environment and an example application under `/example`. The example is deployed here: https://nypl-web-reader.vercel.app.
+The project is build with [esbuild](https://esbuild.github.io/). It uses Typescript, React, Jest and Cypress, and features both a Storybook development environment and an example application under `/example`. The example is deployed here: https://nypl-web-reader.vercel.app.
```
hokei: nit "built"

README.md
```diff
@@ -22,7 +22,7 @@ A big thanks to [R2D2BC](https://github.com/d-i-t-a/R2D2BC) for providing the un
   - [x] Color scheme (night, day, sepia)
   - [x] Fullscreen
   - [x] Paginated / Scrolling mode toggle
-  - [ ] Zoom (PDF only)
+  - [x] Zoom (PDF only)
 - [x] Offline support (prefetch and cache desired content via Service Worker, along with host app shell.
 - [ ] Saving bookmarks / highlights
 - [ ] WAI-ARIA compliant accessibility (_pending accessibility review_)
```
hokei: It has passed accessibility review on our end

cypress/integration/html-reader/settings.ts
```diff
@@ -47,3 +47,28 @@ describe('display settings', () => {
     ); // 4% per step?
   });
 });
+
+describe('useHtmlReader configuration settings', () => {
+  it('should have no injectables by default', () => {
```
hokei: I don't understand - shouldn't R2D2BC throw an error when there are no injectibles?

example/static/samples/test/unparseable-manifst.json
```diff
@@ -0,0 +1 @@
+this is not a valid json file.
```
hokei: nit: typo in filename

cypress/integration/html-reader/settings.ts
kristojorg: No R2 doesn't throw errors if there are no injectables, which is the correct functionality. I recently made a PR so that it throws an error if you pass in an injectable that ultimately fails to load.

hokei (approved):

PR #88, opened 2021-10-19
Add hide and show feature to the page buttons

Toxiapo:
This PR adds the feature to hide the previous button at the start of the book and hides the next button at the end of the book.


src/HtmlReader/index.tsx
```diff
@@ -123,13 +153,26 @@ export default function useHtmlReader(args: ReaderArguments): ReaderReturn {
         autoGeneratePositions: false,
       },
       api: {
-        getContent: getContent as any, //TODO: fix this casting,
-      } as any, //TODO: fix this casting,,
+        getContent: getContent as GetContent,
+        updateCurrentLocation: (location: Locator) => {
+          // This is needed so that setReadingLocation has the updated "reader" value.
+          dispatch({ type: 'LOCATION_CHANGED', location: location });
```
Toxiapo, line 170: @kristojorg Turns out we do need this to trigger an update to the "reader" object for the setReadingLocation function. This makes things much easier to manage!  Instead of calling for an update on multiple places (button click, TOC navigation, scrolling), this method gets triggered on all user actions on the page.

src/PdfReader/index.tsx
```diff
@@ -248,6 +264,50 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
+    } else {
+      const isFirstResource = state.resourceIndex === 0;
+      const isResourceStart = isFirstResource && state.pageNumber === 1;
+
+      const isLastResource =
+        state.resourceIndex === manifest?.readingOrder?.length - 1;
+      const isResourceEnd =
+        (isLastResource && state.pageNumber === state.numPages) ||
+        // On scroll mode, next page button takes you to the next resource. So we can just hide the next button on last resource.
+        (state.isScrolling && isLastResource);
```
Toxiapo, line 268: I think this is the current behavior for webpub pdf page, but I am not sure. @hokei

src/HtmlReader/index.tsx
```diff
@@ -30,7 +36,9 @@ export type HtmlAction =
   | { type: 'SET_SCROLL'; isScrolling: boolean }
   | { type: 'SET_FONT_SIZE'; size: number }
   | { type: 'SET_FONT_FAMILY'; family: FontFamily }
-  | { type: 'SET_CURRENT_TOC_URL'; currentTocUrl: string };
+  | { type: 'SET_CURRENT_TOC_URL'; currentTocUrl: string }
+  | { type: 'LOCATION_CHANGED'; location: Locator }
+  | { type: 'SET_READING_LOCATION'; readingLocation: ReadingLocation };
```
kristojorg: I think I would rename this as it seems a bit confusing with between location and readingLocation. Maybe `LOCATION_CHANGED` and `BOOK_BOUNDARY_CHANGED` and then in state have `location` and `isAtStart` and `isAtEnd`? Idk if those are amazing names or not either but just an idea

src/HtmlReader/index.tsx
```diff
@@ -123,13 +153,26 @@ export default function useHtmlReader(args: ReaderArguments): ReaderReturn {
...
       api: {
-        getContent: getContent as any, //TODO: fix this casting,
-      } as any, //TODO: fix this casting,,
+        getContent: getContent as GetContent,
+        updateCurrentLocation: (location: Locator) => {
+          // This is needed so that setReadingLocation has the updated "reader" value.
+          dispatch({ type: 'LOCATION_CHANGED', location: location });
+          return new Promise(function (resolve, reject) {
+            resolve(location);
+          });
```
kristojorg: I think instead of doing this you should be able to make this function `async` and then just return `location` from it

src/HtmlReader/index.tsx
```diff
@@ -284,3 +327,25 @@ const r2FamilyToFamily: Record<string, FontFamily | undefined> = {
   'sans-serif': 'sans-serif',
   opendyslexic: 'open-dyslexic',
 };
+
+async function setReadingLocation(
+  reader: D2Reader | undefined,
+  dispatch: React.Dispatch<HtmlAction>
+): Promise<void> {
+  if (!reader) return;
```
kristojorg: I don't think this needs to be possible anymore. Should only be called with a `reader`

src/PdfReader/index.tsx
```diff
@@ -248,6 +264,50 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
     }
   }, [containerSize, state.pdfWidth, state.pdfHeight, resizePage]);
 
+  /**
+   * Hide Or Show Page Button
+   */
+  React.useEffect(() => {
+    if (!manifest) return;
+
+    const isSinglePDF = manifest?.readingOrder.length <= 1;
```
kristojorg: I would extract this const out to the top level of the hook, as it seems we will probably be needing it in other places

src/ui/manager.tsx
```diff
@@ -26,16 +26,21 @@ const WebReaderContent: React.FC<ReaderReturn & ReaderManagerArguments> = ({
   ...props
 }) => {
   const bgColor = useColorModeValue('ui.white', 'ui.black', 'ui.sepia');
+  const loading = props.isLoading;
+  const startReading = props.state?.readingLocation.start;
+  const endReading = props.state?.readingLocation.end;
```
kristojorg: I would rename these to `isAtStart` and `isAtEnd` for readability. They currently seem more like actions than states based on the name

kristojorg (changes requested):
This looks great (!), just a few comments/requests:

I am not really able to test the single page PDF version because the example we have is so large that it slows my browser to a stall in scrolling mode. Especially bad in safari, but also in chrome I can't get it to switch back and forth from paginated to scrolling and back. @hokei do you know about these perf problems on the long scrolling pdf? Should we add a shorter one for dev purposes for now?

Finally, can we add tests for this? I imagine the following things we want to verify:

- Doesn't show on initial load
- Does show when you click next page (or scroll away if we can test that)
- Hides the next page button when you get to the last page
- Going back from last page re-shows the next page button
- Switching between scrolling/paginated doesn't mess this up

hokei:
Yes, I'll switch out the PDF, this has been an issue.  
The perf problems do need to be resolved - I know why they're happening but to fix them would be quite an ordeal.  The single page pdf example shouldn't have any next/previous buttons at all, and that's what I'm seeing.  

There should be a test that a single resource webpub in scrolling mode doesn't have page arrows at all.

src/PdfReader/index.tsx
hokei: Yes, this is correct.

hokei (commented):
Seems to not work in a tricky edge case:  
1. Load an epub and navigate to the last page.  
2. Resize your window so that the entirety of the last page doesn't fit on the epub any more.  
3. There's no way to get to the end of the epub now (though if you navigate to the previous page you can then navigate back).  

Is this a quick fix?  If not we should file it as a bug since it's pretty esoteric.  
We should have a test for something like this though.

Toxiapo:
> Seems to not work in a tricky edge case:
> 
> 1. Load an epub and navigate to the last page.
> 2. Resize your window so that the entirety of the last page doesn't fit on the epub any more.
> 3. There's no way to get to the end of the epub now (though if you navigate to the previous page you can then navigate back).
> 
> Is this a quick fix? If not we should file it as a bug since it's pretty esoteric. We should have a test for something like this though.

Yes, I can add a resize event to the page for the fix. However, I think the updateCurrentLocation function from the R2builder should handle the resizing behavior but isn't..

cypress/integration/multi-pdf/page-button.ts
```diff
@@ -0,0 +1,54 @@
...
+    cy.findByRole('menuitem', { name: 'Index' }).click();
+
+    cy.wait('@pdf');
+    cy.wait(1000);
+
+    // FIXME: Find a smaller PDF or a better way to simulate 10 clicks
+    let clicks = 10;
+    while (clicks-- > 0) {
+      cy.findByRole('button', { name: 'Next Page' }).click();
+    }
```
kristojorg, line 42: I'm surprised this works doing it in such quick succession like this

src/HtmlReader/index.tsx
```diff
@@ -141,15 +162,33 @@ export default function useHtmlReader(args: ReaderArguments): ReaderReturn {
...
       dispatch({ type: 'SET_READER', reader });
+      enableResizeEvent(reader, dispatch);
     });
-  }, [webpubManifestUrl, getContent, injectables, injectablesFixed]);
+  }, [
+    webpubManifestUrl,
+    getContent,
+    injectables,
+    injectablesFixed,
+    state.isScrolling,
```
kristojorg: I don't think we want to re-initialize the reader every time the `state.isScrolling` state changes. It should already be updated when it changes. If you put it in here so you could use the default value I think you should bring only the default in to this effect, not the state variable

kristojorg (changes requested):
Just the one thing but otherwise looks good!

cypress/integration/multi-pdf/page-button.ts
Toxiapo: Yea, I think it was because this action doesn't trigger a state change to the page button nor make a new request, so the content switch was instant.

src/HtmlReader/index.tsx
Toxiapo: Updated, thank you for catching this!

kristojorg (approved):
Looks good!

PR #89, opened 2021-10-20
Chore/paralell cypress

kristojorg:
This update will run 6 containers with cypress to make these runs faster. Times are dropping from about 4:30 -> 2:00 from what I can tell. We should do the same in CPW if this proves successful


.github/workflows/cypress.yml
```diff
@@ -1,19 +1,77 @@
...
+      #     if-no-files-found: error
+      #     path: build
+
+      # - name: Cypress install
+      #   uses: cypress-io/github-action@v2
+      #   with:
+      #     # Disable running of tests within install job
+      #     runTests: false
+      #     # build: yarn build
+      #     install-command: npm i
```
samanthaandrews: can we remove this? otherwise, looks good!

samanthaandrews (changes requested):

PR #93, opened 2021-10-20
Fixing TOC icon alignment by using `leftIcon` prop in Chakra <MenuButton />

samanthaandrews:
Issue: TOC icon was out of alignment on https://qa-web.openebooks.us/

Screen Shot 2021-10-20 at 1 33 14 PM

Fix: The <span /> surrounding the icon and text needed the following css:

```
span { 
    align-items: center;
    display: inline-flex;
}
```

I fixed this by putting the icons into the `leftIcon` prop in the chakra <MenuButton />. I _think_ this solves the issue. I played around with a CodePen replication of it, since it's hard to know how it will actually behave in openebooks. You can check it out here: [https://codepen.io/samandrews/pen/WNEGOJP](https://codepen.io/samandrews/pen/WNEGOJP)

Final result looks the same in web-reader:

Screen Shot 2021-10-20 at 2 08 15 PM


Toxiapo (approved):
LGTM!

PR #96, opened 2021-10-21
Add accessibility check

Toxiapo:


tests/utils/MockData.tsx
```diff
@@ -149,25 +151,29 @@ export const MockReaderState = {} as ReaderState;
 
 const MockComponent = (): React.ReactElement => <>Hello, world.</>;
 
-const MockHtmlReaderState = {
-  colorMode: 'day',
+export const MockHtmlReaderState = {
+  colorMode: 'day' as ColorMode,
```
kristojorg, line 155: This is fine, but just so you know for the future, you can use `colorMode: 'day' as const` to restrict it from any `string` to just the string `'day'`, which would pass the later typecheck. You could do the same for the whole `MockHtmlReaderState` even if you wanted

kristojorg (approved):
Looks good to me! LGTM

tests/utils/MockData.tsx
Toxiapo: Got it, thank you!

PR #99, opened 2021-10-21
Fix injectables tests

samanthaandrews:
Switching to use the Alice Epub example in these tests to fix the `parent element not found` error


kristojorg (approved):
Weird that it's book specific, but good to know.

PR #100, opened 2021-10-22
SFR_1363 - Fix the encrypted example

kristojorg:
We were previously not properly checking for encryption info in the example when decrypting a book. This meant that if one file was unencrypted (as was the case with the TOC file in the book we used), we would try to decrypt it and get errors. I'd already fixed this actually in the `getContent` function in CPW, but not in the example app here.


hokei (commented):
I'm trying to understand what's going on here - when you say this is fixed in cpw, does this mean that this is all duplicated in CPW and webreader?  The logic changes in here make it seem like this is all necessary and should live here, rather than CPW, but also the comment makes it seem like this was just fixing the example, and that CPW is working.  Can you point me to where the CPW fixed happened so I can figure out what's happening now that wasn't happening earlier?

kristojorg:
Basically yes, there is duplicate code between the web-reader __example app__ and CPW. This is specifically the `getContent` function that is passed in to the web-reader to fully decrypt AxisNow books:
- Get a `ReadiumLink` and decide whether it should be decrypted (based on the info in the link).
- Fetch the asset
- Use the decryptor to decrypt it
- Search for any embedded assets (images etc), properly resolve their hrefs, decrypt them if necessary, and inline the decrypted content into the original html document.

All of this functionality is specific to axisnow books. I don't think it should be included in the web-reader code. However, for ease of development I included an AxisNow example in the example app, and thus need to use this logic in the example app as well as the web-reader. 

It was fixed in CPW when I first added this code to CPW. I copied the code in to CPW and then noticed a couple problems and immediately fixed them, but didn't port them back to the web-reader at the time. This PR is basically just doing that (and also fixes one issue that happens with locally-hosted axisnow books which will never happen in CPW because they're all remote).

So anyways, I don't think this code should be exported from this repo. If anything it should live in the axisnow decryptor package. I think I'll try to add it to there today actually.

kristojorg:
I made this pr to move the logic to the decryptor package: https://github.com/NYPL-Simplified-Packages/axisnow-access-control-web/pull/8

hokei (approved):

PR #101, opened 2021-10-25
Fix scrollbars

kristojorg:
This PR will hopefully make scrollbars better. It moves the page buttons to the bottom of the page, and also allows for the web-reader to be embedded in a page.


Toxiapo (changes requested):
There seems to be a bug on PDF pages.

If you visit a PDF page and switch to Scrolling mode, the Footer gets pushed all the way down to the bottom of the content, and if you scroll the page, the Header does not stay on the screen.

https://web-reader-git-fix-scrollers-nypl.vercel.app/pdf-collection

src/PdfReader/index.tsx
```diff
@@ -424,6 +428,16 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
         id={IFRAME_WRAPPER_ID}
         ref={containerRef}
       >
+        {/* FIXME: POC, update this with more react proach*/}
+        <style>
+          {`
+            .react-pdf__Document {
+              height: calc(100vh - ${HEADER_HEIGHT + FOOTER_HEIGHT}px);
```
kristojorg, line 484: I think this isn't going to work because in the embedded example we are no longer assuming the pdf is full screen

cypress/support/pdfZoomTestHelper.ts
```diff
@@ -0,0 +1,13 @@
+export const pdfZoomTestHelper = (
+  $elm: JQuery<HTMLElement>,
```
hokei, line 2: Just interested in what this selector is, is it related to the JQuery of yore?  (predictably, this is hard to google haha)

kristojorg: Hmmm that is a good question, it's something @Toxiapo added, maybe he can clarify

Toxiapo: Yup, it's the good old Jquery object. I was using the [cy.get](https://docs.cypress.io/api/commands/get#Syntax) method to get the pdf page element, it returns a Jquery object in its callback.

cypress/integration/html-reader/toc-navigation.ts
```diff
@@ -66,7 +66,7 @@ describe('navigating an EPUB page', () => {
     cy.findByRole('button', { name: 'Table of Contents' }).click();
 
     cy.log('open chapter 1');
-    cy.findByRole('menuitem', { name: 'I: Down the Rab­bit-Hole' }).click();
+    cy.findByRole('listitem', { name: 'I: Down the Rab­bit-Hole' }).click();
```
hokei: I'm going to be annoying and ask about this stuff now - was `menuitem` being flaky?

cypress/integration/multi-pdf/navigations.ts
```diff
@@ -7,13 +7,15 @@ describe('Multi PDF navigation', () => {
...
     cy.findByRole('button', { name: 'Table of Contents' }).click();
-    cy.findByRole('menuitem', { name: 'Foreword' }).click();
+    cy.findByRole('listitem', { name: 'Foreword' }).click();
     cy.wait(5000);
     cy.findByText('Anthropology without Informants', { timeout: 10000 }).should(
       'be.visible'
     );
     cy.findByRole('button', { name: 'Next Page' }).click();
-    cy.findByText('Foreword', { timeout: 10000 }).should('exist');
+    cy.get('div[class="react-pdf__Page__textContent"]').within(() => {
```
hokei: In RTL philosophy we don't normally use class selectors - is this ok because we don't expect `react-pdf` classnames to change, or because there's really no way to get this content with role selectors, or both?

hokei:
In embedded mode, should the overlay cover all of the page, or just the reader insert?  (Probably new ticket)

cypress/integration/multi-pdf/navigations.ts
```diff
@@ -22,8 +24,8 @@ describe('Multi PDF navigation', () => {
     cy.findByRole('radio', { name: 'Paginated' }).should('not.be.checked');
     cy.findByRole('radio', { name: 'Scrolling' }).should('be.checked');
     cy.findByRole('button', { name: 'Table of Contents' }).click();
-    cy.findByRole('menuitem', { name: 'Foreword' }).click();
-    cy.wait(5000);
+    cy.findByRole('listitem', { name: 'Foreword' }).click();
+    cy.wait('@pdf');
```
hokei, line 29: So this @pdf refers to the line in `/support/commands.ts` How do we decide what is a command and what is a util?

hokei: eg: Why is /util/pdfZoomTestHelper not an additional command, instead of being a util function?

src/constants.ts
```diff
@@ -1,3 +1,11 @@
 export const ReadiumWebpubContext = 'http://readium.org/webpub/default.jsonld';
 
 export const IS_DEV = process.env.NODE_ENV === 'development';
+
+// we have to set a constant height to make this work with R2D2BC
+export const HEADER_HEIGHT = 48;
```
hokei, line 6: new ticket:  How do we make this stuff configurable?

src/ui/Header.tsx
```diff
@@ -38,20 +38,23 @@ export const DefaultHeaderLeft = (): React.ReactElement => {
...
+  const { headerLeft, navigator, manifest } = props;
   const mainBgColor = useColorModeValue('ui.white', 'ui.black', 'ui.sepia');
 
   return (
-    <HeaderWrapper bgColor={mainBgColor}>
+    <HeaderWrapper bg={mainBgColor}>
       {headerLeft ?? <DefaultHeaderLeft />}
       <HStack ml="auto" spacing={1}>
         <TableOfContent
+          containerRef={props.containerRef}
```
hokei: nit: assign containerRef in L47?

src/ui/Header.tsx
```diff
@@ -90,9 +94,9 @@ export const HeaderWrapper: React.FC<ComponentProps<typeof Flex>> = ({
       px={8}
       borderBottom="1px solid"
       borderColor="gray.100"
-      backgroundColor={bgColor}
+      {...rest}
```
hokei, line 97: Do we know exactly what might be passed into HeaderWrapper?  If so, would it make sense to just specify them?

src/ui/TableOfContent.tsx
```diff
@@ -1,40 +1,37 @@
...
 import {
   Link,
-  Menu,
-  MenuButton,
-  MenuItem,
-  MenuList,
   Portal,
   Text,
   Icon,
+  useDisclosure,
```
hokei: this is awesome

src/PdfReader/index.tsx
```diff
@@ -42,6 +47,7 @@ type PdfReaderAction =
     };
 
 const IFRAME_WRAPPER_ID = 'iframe-wrapper';
+export const SCALE_STEP = 0.1;
```
hokei, line 50: nit: should this be in `constants` ?

src/PdfReader/index.tsx
hokei: This looks like it works now?

hokei:
growing-height-embedded-pdf 404s

hokei:
How can you navigate the table of contents in the embedded example?  Seems [like](https://github.com/NYPL-Simplified/web-reader/pull/101/files#diff-d99fe88bf20d43ca89f5672fa0eaf005cdd2be15fd70c31fa91608857aa01c0dL22) this continues to work so I'm not sure why it would work sometimes and not others?

cypress/integration/html-reader/toc-navigation.ts
kristojorg: No I changed the html to render list items instead of menuitems. I think that's more semantic anyways for this use case

cypress/integration/multi-pdf/navigations.ts
kristojorg: @Toxiapo this one for you : )

src/PdfReader/index.tsx
kristojorg: Good question. I think it's specific to this file so fine to leave in here for now personally

src/ui/TableOfContent.tsx
kristojorg: Yea chakra knows what's up :)

src/ui/Header.tsx
kristojorg: I'm basically just trying to allow any props that might be passed to the div element be passed along. This allows the consumer to use style props on it: `<HeaderWrapper bg="red">` or some other style prop as necessary. For now it looks like we only use `bg` but I basically wanted `<HeaderWrapper>` to function like a `<Flex>` except with some default styles.

cypress/integration/multi-pdf/navigations.ts
kristojorg: Another q for @Toxiapo

kristojorg:
> In embedded mode, should the overlay cover all of the page, or just the reader insert? (Probably new ticket)

Do you mean the loading indicator overlay? yea that should only cover the embedded thing but we don't control it right now as it's part of R2D2BC.

> How can you navigate the table of contents in the embedded example? Seems like this continues to work so I'm not sure why it would work sometimes and not others?

I am able to click it and click on items as usual. What are you seeing?

> growing-height-embedded-pdf 404s

Fixed, thanks

src/constants.ts
kristojorg: Yea good point for a new ticket. Wonder where I should put that one

cypress/integration/multi-pdf/navigations.ts
Toxiapo: Yeah there's only a single instance of `react-pdf__Page__textContent` so this should be fine, the reason that it's currently wrapped with `.within` was that the cypress was complaining about 'multiple matches', which make sense now that the TOC is always on the page.

I think the better test is to test a paragraph instead of just the title of the page. Will post an update now!

src/constants.ts
hokei: Probably some kind of config object?  I was thinking about this when trying to make `defaultScrolling` configurable - we'll need to figure out a way have the app consume something like this, cuz I don't think it's tenable to put them in props forever.

hokei:
> I am able to click it and click on items as usual. What are you seeing? 
sorry i mean with the keyboard.  This should probably be automated-tested as well

cypress/integration/multi-pdf/navigations.ts
Toxiapo: I think both works. I would define a command for this if this was being used elsewhere, but this is only testing the zoom value specifically.

kristojorg:
@hokei I added a ticket to add the TOC tests: https://jira.nypl.org/browse/SFR-1396

Toxiapo (approved):
Looks Great! Love the new footer! 🎉

PR #102, opened 2021-10-27
SFR-1383 paginate by default

hokei:
Starts the web reader in pagination mode 
Exposes settings so that a user can control which mode the reader starts in


src/PdfReader/index.tsx
```diff
@@ -377,8 +400,10 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
       navigator: {
         goForward,
         goBackward,
-        zoomIn,
-        zoomOut,
+        increaseFontSize,
+        decreaseFontSize,
+        setFontFamily,
+        setColorMode,
```
samanthaandrews: Should these be in the PdfReader?

src/ui/Header.tsx
```diff
@@ -89,6 +89,7 @@ export const HeaderWrapper: React.FC<ComponentProps<typeof Flex>> = ({
       px={8}
       borderBottom="1px solid"
       borderColor="gray.100"
+      backgroundColor="ui.white"
```
samanthaandrews: i noticed that in scroll mode you could see the text behind the header. making header background white fixes it.

Toxiapo: We do want to support color mode here, so you might want to pass down the `bgColor` prop from the header

src/HtmlReader/index.tsx
```diff
@@ -137,7 +149,13 @@ export default function useHtmlReader(args: ReaderArguments): ReaderReturn {
     }).then((reader) => {
       dispatch({ type: 'SET_READER', reader });
     });
-  }, [webpubManifestUrl, getContent, injectables, injectablesFixed]);
+  }, [
+    webpubManifestUrl,
+    getContent,
+    injectables,
+    injectablesFixed,
+    readerSettings,
```
Toxiapo: stray `readerSettings`

src/ui/Header.tsx
samanthaandrews: good catch! fixed now

src/HtmlReader/index.tsx
samanthaandrews: i think this is meant to be here, but i'll let @hokei confirm

hokei: I think this can be taken out, as it is not used.

samanthaandrews (approved):

PR #103, opened 2021-10-27
SFR-1372 Fix slow PDF rendering

samanthaandrews:
Improves slow single-resource PDF rendering by introducing the `useInView()` hook from `react-intersection-observer` which tells us when an element is in view and only renders pdf pages if a user has scrolled to that element.

I attempted to optimistically render the next page before it is actually in view, but encountered some challenges when attempting to store inView booleans in an array state variable, and ultimately I'm not sure it is possible to do with this `useInView` hook. Lmk if you want to pair on this more to try to understand it more in-depth!


src/PdfReader/index.tsx
```diff
@@ -429,11 +418,14 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
             <>
               {state.isScrolling &&
                 Array.from(new Array(state.numPages), (_, index) => (
-                  <ChakraPage
+                  <ScrollPage
                     key={`page_${index + 1}`}
                     width={containerSize?.width}
```
samanthaandrews, line 421: I want to clear up the difference between container size and pdf height and width

src/PdfReader/ScrollPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+  scale,
+  index,
+  width,
+  onLoadSuccess,
+  placeholderHeight,
+  placeholderWidth,
+}) => {
+  const { ref, inView } = useInView({
+    threshold: 0,
+    triggerOnce: true,
```
samanthaandrews, line 41: This option makes it so that pages that have been rendered before don't un-render.

kristojorg:
This works crazy good, LGTM!

kristojorg (approved):

PR #104, opened 2021-10-29
Define a  default dimension object for the useMeasure hooks instead of null

Toxiapo:
[Jira Ticket](https://jira.nypl.org/browse/SFR-1387)

There's a bug on the PDF page where when the PDF is still loading the content, and immediately tab to a different page, the PDF would fail and return an error.

The logic failed from the `containerSize` variable where it was defaulted to be null, however, adding a default value to the `containerSize` seems to fix the issue. The pdf hook re-renders itself whenever the content updates, we always get the latest container dimensions. 

@kristojorg I'd like your input on this and see if this makes sense to do it this way.


kristojorg (approved):
This seems fine to me, especially if it resolves the issue. Nice!

PR #105, opened 2021-11-02
PDF: disable page navigation on scroll mode

Toxiapo:
This PR addresses the inoperable page button bug on scrolling mode, it happens because we are trying to navigate between pages while it currently only supports navigation between resources on the webpub PDF pages with multiple resources.


hokei:
Trying to figure out what this is doing - if you hit "single pdf webpub" you still see the page navigation buttons in scroll mode.  If you hit those buttons, they don't do anything.

Toxiapo:
> Trying to figure out what this is doing - if you hit "single pdf webpub" you still see the page navigation buttons in scroll mode. If you hit those buttons, they don't do anything.

On pdf with multiple resources, the page buttons on scrolling mode should navigate users between resources, I think this is the setup we have now.

https://user-images.githubusercontent.com/14434233/140088441-5d31146f-5fa8-4c0b-80f1-0f5099978569.mov

hokei (changes requested):
Ah!  Looks good, please add a test for this behavior so we can catch it if it regresses.

hokei (approved):
LGTM!  In the future esp if I'm the only reviewer and it seems like something's fallen off like this please don't hesitate to ping me.

Toxiapo:
> LGTM! In the future esp if I'm the only reviewer and it seems like something's fallen off like this please don't hesitate to ping me.

I will make sure to ping next time, thanks!

PR #106, opened 2021-11-19
Show missing TOC message

Toxiapo:
This PR adds a Missing TOC message to users when there's no TOC supplied in the manifest.

### DIFF
--- a/src/HtmlReader/HtmlReaderContent.tsx
+++ b/src/HtmlReader/HtmlReaderContent.tsx
@@ -5,7 +5,7 @@ const HtmlReaderContent: React.FC<{
   isScrolling: boolean;
   growsWhenScrolling: boolean;
 }> = ({ height, isScrolling, growsWhenScrolling }) => {
-  const shouldGrow = isScrolling && growsWhenScrolling;
+  // const shouldGrow = isScrolling && growsWhenScrolling;
   return (
     <div id="D2Reader-Container">
       <main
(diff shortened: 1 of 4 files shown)


src/HtmlReader/HtmlReaderContent.tsx
```diff
@@ -19,7 +19,7 @@ const HtmlReaderContent: React.FC<{
            * content. In paginated mode, however, we must have this set because
            * we have to decide how big the content should be.
            */
-          height: shouldGrow ? 'initial' : height,
```
Toxiapo: I am not sure if this is needed anymore? When I am on scrolling mode with a big page and click on the TOC button, the popup would just jump to the top of the page blocking the header. (e.b /moby-epub2)

samanthaandrews (approved):
LGTM!

src/HtmlReader/HtmlReaderContent.tsx
hokei: I'm not sure if this is needed, but if it isn't, I think L8 should be removed rather than commented out.

src/HtmlReader/HtmlReaderContent.tsx
```diff
@@ -4,22 +4,14 @@ const HtmlReaderContent: React.FC<{
...
         id="iframe-wrapper"
         style={{
-          /**
-           * This determines the height of the iframe.
-           *
-           * If we remove this, then in scrolling mode it simply grows to fit
-           * content. In paginated mode, however, we must have this set because
-           * we have to decide how big the content should be.
-           */
-          height: shouldGrow ? 'initial' : height,
```
kristojorg, line 22: Hang on, did we intend to remove this?

kristojorg: We need to put this back actually. Maybe my comment wasn't clear what the use-case is for it, but this specifically affects growing-height readers. So go to the growing height embedded moby dick, and change it to scrolling mode. Then you will see that it does not actually grow in height anymore

PR #110, opened 2021-12-08
Upgrade r2d2bc

Toxiapo:
This upgrade fixes [OE-227](https://jira.nypl.org/browse/OE-227), that, when the user clicks on an internal link, it should navigate the user if it's found in the readeringOrder in the manifest instead of downloading the resource.

### DIFF
--- a/cypress/integration/html-reader/internal-link.ts
+++ b/cypress/integration/html-reader/internal-link.ts
@@ -8,9 +8,6 @@ describe('navigating an EPUB page using internal link', () => {
     });
 
     cy.loadPage('/moby-epub2');
-
-    // Use a consistent viewport
-    cy.viewport(1024, 768);
   });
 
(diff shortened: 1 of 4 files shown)


hokei:
Considering that r2d2bc has no tests, can we add one here for it, or at least have an example of a book that shows this functionality?

Toxiapo:
> Considering that r2d2bc has no tests, can we add one here for it, or at least have an example of a book that shows this functionality?

Oh right, good point, I will add one in.

kristojorg (approved):
looks good!

kristojorg:
Although the cypress tests are passing : /

Toxiapo:
> Although the cypress tests are passing : /

ahhhh, it's the random `Cannot read properties of null (reading 'ownerDocument')` error from the reader..

Issue #115, opened 2021-12-22
skipWaiting for SW to be installed


AntonOfTheWoods:
According to my reading (https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil) and tests

https://github.com/NYPL-Simplified/web-reader/blob/e7ea14c7e12bce49817df4d9becaefe539b51420/src/ServiceWorker/sw.ts#L43

Requires a thenable, not a function. Should it not be

```
event.waitUntil(installSW());
```

?

Toxiapo:
`event.waitUntil(installSW); ` should be fine with the way `installSW` was written, it's an async function without an inner function.

The caching does not happen on the initial install but after the client fetches the manifests and other resources. https://github.com/NYPL-Simplified/web-reader/blob/e7ea14c7e12bce49817df4d9becaefe539b51420/src/ServiceWorker/sw.ts#L50

Is there any error you are seeing with this function? I am happy to debug the issue together.

AntonOfTheWoods:
Thanks @Toxiapo . Did you actually test it? For me it simply never executes at all without the parentheses. My understanding of the method signature suggests that would also be the expected behaviour (or an error). The method expects a promise, not a method as a parameter. Being an async method, it returns a promise only if it is actually executed.

If you do try and test, you would need to check whether you actually get a SW waiting or not, as the log method doesn't print when the build is production. Everything works as expected with parentheses, it doesn't work at all for me without.

Toxiapo:
@AntonOfTheWoods Ahh good catch! You are right, the function did not work for me without the parentheses!!

Feel free to make a PR with the fix! : )

PR #116, opened 2021-12-23
Fix missing parentheses in call to skipWaiting

AntonOfTheWoods:
The intention of the code is to not wait for the pages the current
service worker is controlling to be closed before updating the SW. This
was not happening because an async method was being passed to waitUntil,
rather than actually calling it and thus producing the promise that this
method expects in the SW.

I am unsure how this might be tested (it obviously wasn't previously!), and definitely have no relevant experience with doing so in a robust manner. Please let me know if there is a good way to test this.

### DIFF
--- a/src/ServiceWorker/sw.ts
+++ b/src/ServiceWorker/sw.ts
@@ -40,7 +40,7 @@ export default function initWebReaderSW({
       await self.skipWaiting();
       log('INSTALLED');
     }
-    event.waitUntil(installSW);
+    event.waitUntil(installSW());
   });
 
   /**


AntonOfTheWoods:
Fixes https://github.com/NYPL-Simplified/web-reader/issues/115#issuecomment-999990793

Toxiapo:
Thank you for the PR, @AntonOfTheWoods! This looks good to me.

If you want, you could write some unit tests for it to make sure the activation, caching, and clearing cache works using something like [service-worker-mock](https://github.com/zackargyle/service-workers/tree/master/packages/service-worker-mock), up to you!

kristojorg:
@AntonOfTheWoods I'm going to go ahead and approve this, as you're right that this was a necessary change. 

With that said, we are currently _not_ using the publication service worker in our projects because it turned out that precaching the publication resources was actually blocking the main thread (or was slowing down overall network speed, I'm not sure). The end result was that books took far longer to load if we were trying to precache their resources at the same time. I don't see this happening in the example app, possibly because these resources are much lighter weight, but I was seeing it in our projects.

We do have an intention to resolve this, probably by moving to another precache mechanism such as `<link rel="prefetch">` either in the host window's `<head>` or in the iframe's `<head>`. I had originally expected service worker precaching to be de-prioritized (which is what resources online say), but that doesn't seem to the be the case here. Anyway, I just thought I'd let you know about that issue.

kristojorg (approved):

Toxiapo (approved):

PR #118, opened 2022-01-28
Update packages and enforce node version

Toxiapo:
This PR updates some of the package versions that would cause build errors when building the project from scratch and enforces a node version with a .nvmrc file.


package.json
```diff
@@ -144,8 +144,8 @@
     "typescript": "^4.3.5"
   },
   "dependencies": {
-    "@chakra-ui/react": "^1.6.9",
-    "@d-i-t-a/reader": "^2.0.0-beta.6",
+    "@chakra-ui/react": "1.6.9",
+    "@d-i-t-a/reader": "2.0.0-beta.6",
```
Toxiapo, line 148: Keeping these two fixed versions because the new chakraui version is causing a build error on the Text component and version >2.0.0-beta.6 causes parcel build error...

.nvmrc
```diff
@@ -0,0 +1 @@
+v16.3.0
```
kristojorg: I think we should make this 14.9.0. AWS doesn't support node 16 yet in serverless functions

kristojorg (changes requested):

kristojorg (approved):
LGTM

PR #119, opened 2022-02-07
changed version number to v1

hokei:


.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
...
+        id: create_release
+        if: steps.autotagger.outputs.tagname != ''
+        uses: actions/create-release@v1.0.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ steps.autotagger.outputs.tagname }}
+          release_name: ${{ steps.autotagger.outputs.tagname }}
+          # use the body of the PR commit as the release body
+          body: "$commitmsg"
```
kristojorg: Our current release process using `npm run release` creates a release on github including bullets for every PR since the last release, and linking to those commits. I quite like that feature and would like to find a way to keep something like it if we move to automated releases. This current setup will use whatever is in the body of the release commit. 

For the time being, tbh, I think it's fine to keep the manual releases unless we want to set up separate `dev` and `release` branches. I don't really see a need to do that though

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nypl/web-reader",
-  "version": "0.3.3",
+  "version": "1.0.0",
```
kristojorg, line 3: I think we should stay below 1.0.0 until the remove R2D2BC PR is merged. It doesn't actually break any APIs or features, but it's a pretty major overhaul and I think it'd be good to wait on the 1.0. What do you think?

hokei: We talked about this in DRB, and we can all get on a call and talk about this - but the web reader should have been in V1 when it released in October.  If the R2D2BC stuff causes breaking changes, it should be V2.  This kind of makes sense - the web reader has been in production, stably, for some months.  Really it's semantic and if you feel that it's not actually yet "released", that's fine.  We were just getting questions as to "when is web reader released" and the answer that felt truest to me was that it was "released" in October but we forgot to bump the number then.  We then made a lot of changes to it, so removing R2D2BC is at the very least a new feature.

.github/workflows/release.yml
hokei: The only thing I want to make sure of is that this is something that more than one person has npm permissions to do.  But the point of the PR was to get a V1 released, and I was only going to do this if it was uncontroversial.  If we want to actually talk about it let's make a new ticket to talk about how it would work.

kristojorg (approved):
Looks good!

PR #91, opened 2021-10-20
With custom HTML Renderer

kristojorg:
This is a draft PR showing what it would look like to remove R2D2BC.

TODO:

- [x] Color mode
- [x] Font size
- [x] Font family
- [x] Paginated mode
- [x] Snap to page in paginated mode
- [x] TOC navigation
- [x] Internal navigation links
- [x] Make internal link clicks register a change of location so prev/next page work
- [x] fix scroll bars
- [x] Remember where user is when switching between paginated and scroll mode
- [x] Scroll to end of resource when navigating backwards
- [x] Keep location in URL bar
- [x] paginated window resize
- [x] Fix bug when there is a "half page"

PR #121, opened 2022-02-17
Add extra cypress tests

Toxiapo:
This PR adds extra cypress tests to the HTML renderer section and also validates other tests, which should cover the following test scenarios

- [x] when you scroll to the bottom of the chapter and press "next", it knows that it's at the end of the chapter and loads the next one
- [x] when you scroll, then resize, it knows you're not yet at the end of the chapter
- [x] Internal links works on both paginated and scrolling mode
- [x] TOC navigates on paginated and scrolling mode
- [x] Height and width problem, make sure there are no extra scrollbars
- [x] All the settings under the settings button
- [x] Next / Previous button response to book state (navigates + hide/show state)

### DIFF
--- a/cypress/integration/html-reader/page-button.ts
+++ b/cypress/integration/html-reader/page-button.ts
@@ -1,5 +1,3 @@
-import { IFRAME_SELECTOR } from '../../support/constants';
-
 describe('PageButton visibility on useHtmlReader', () => {
   beforeEach(() => {
     cy.loadPage('/streamed-alice-epub');
@@ -43,7 +41,7 @@ describe('PageButton visibility on useHtmlReader', () => {
 
     cy.wait(1000);
 
-    cy.getIframeBody(IFRAME_SELECTOR).find('.copyright-page').should('exist');
+    cy.getIframeBody().find('.copyright-page').should('exist');
 
     cy.wait(1000);
 
@@ -79,7 +77,7 @@ describe('PageButton visibility on useHtmlReader', () => {
 
     cy.wait(1000);
 
-    cy.getIframeBody(IFRAME_SELECTOR).find('.copyright-page').should('exist');
+    cy.getIframeBody().find('.copyright-page').should('exist');
 
     cy.wait(1000);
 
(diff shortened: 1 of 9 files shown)


samanthaandrews (approved):
nice!

cypress/integration/html-reader/page-button.ts
```diff
@@ -119,4 +138,42 @@ describe('PageButton visibility on useHtmlReader', () => {
...
+      'not.be.disabled'
+    );
+
+    cy.log('Small screen should disable next button');
+    cy.viewport(100, 100);
+
+    cy.findByRole('button', { name: 'Next Page' }).should('be.disabled');
+    cy.findByRole('button', { name: 'Previous Page' }).should(
+      'not.be.disabled'
+    );
```
kristojorg, line 177: I don't think this is testing what we want. We want to test that screen resize will trigger an update on the disabled state of the prev/next buttons. In this case, they stay the same.

kristojorg (changes requested):
This looks great, just the one test I think we should alter a bit!

cypress/integration/html-reader/page-button.ts
Toxiapo: Hmm okay, this test is just not correct actually, next page button should always be disabled on scrolling mode cuz it's on the last page(For the current webreader version, it's chapter navigation). This is only true for paginated mode where a resize will trigger the next/previous state, we already have a test for that.  

Butttt, I think this is different in the new HTML renderer? I can make a comment to change this after the merge if this is what we decided to do.

kristojorg: I think the new renderer and the old will function the same. When in scrolling mode, the next/prev button should take you one page unit down the page in scrolling mode (as well as paginated mode)

Toxiapo: I think we might have found a bug. It's chapter navigation for me on the current version.

I am going to file a ticket for this bug to fix this and monitor the bug after we merge #91

Toxiapo: Ticket: [OE-376](https://jira.nypl.org/browse/OE-376)

PR #125, opened 2022-03-23
OE 340/save reading position

kristojorg:
This saves the reading position and current settings to local storage. Note that we do not do this for the PDF reader for the time being. I think in order to enable that we want to re-think how these reducers are being used and modularize some of their code a bit. 

One note: I did more all settings into a `settings` object

### DIFF
--- a/cypress/integration/html-reader/location-sync.ts
+++ b/cypress/integration/html-reader/location-sync.ts
@@ -7,7 +7,7 @@ it('should remember location between page visits', () => {
 
   // navigate to some page
   cy.findByRole('button', { name: 'Table of Contents' }).click();
-  cy.wait(100);
+  cy.wait(1000);
   cy.findByRole('menuitem', { name: 'Chapter 10. A Bosom Friend.' }).click();
   cy.wait(1000);
 
@@ -21,7 +21,7 @@ it('should remember location between page visits', () => {
   cy.findByRole('link', { name: 'Moby Dick' }).click();
   cy.wait(1000);
   cy.findByRole('button', { name: 'Table of Contents' }).click();
-  cy.wait(100);
+  cy.wait(1000);
   cy.findByRole('menuitem', { name: 'CHAPTER 5. Breakfast.' }).click();
   cy.getIframeBody()
     .find('h2')
--- a/cypress/integration/html-reader/location-sync.ts
+++ b/cypress/integration/html-reader/location-sync.ts
@@ -1,0 +1,49 @@
+it('should remember location between page visits', () => {
+  cy.visit('/html/moby-epub3', {
+    onBeforeLoad: (win) => {
+      win.localStorage.clear();
+    },
+  });
+
+  // navigate to some page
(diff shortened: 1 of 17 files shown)


Toxiapo (commented):
Tested on Chrome/Safari/FF, all looks great!

@kristojorg  For this PR, we are not saving the user settings as part of the URL query, right? (Just to confirm cuz you have them saved in the local storage.)

Toxiapo (approved):

kristojorg:
> For this PR, we are not saving the user settings as part of the URL query, right? (Just to confirm cuz you have them saved in the local storage.)

That's correct. Putting the location in URL query is mostly for ability to share a page between users or bookmark a page, but I didn't think settings should be shared/stored in the same way.

kristojorg:
It looks like the cypress PDF tests are failing to fetch the PDF due to CORS issues. I'm confused how this worked previously though. @hokei do you have any idea how this is supposed to work? Is the example meant to be run with a CORS_PROXY_URL env var? I didn't change anything there so not sure why it would have stopped working now

hokei:
>  Is the example meant to be run with a CORS_PROXY_URL env var? I didn't change anything there so not sure why it would have stopped working now

I don't know why not touching it would have caused anything to break but it looks like the PDFs are loading resources from academic publishers and a lot of them do lock down their CORS settings, so they wouldn't load unless you have the proxy set.  This should be available for the webreader though, since the vercel deployed version is using it.  is it set in the env file that Cypress is using?

hokei:
Alternately,  you can use a self-hosted PDF in examples for testing, this seems a lot less brittle but seems out of scope for this ticket.

cypress/integration/html-reader/location-sync.ts
```diff
@@ -16,9 +16,8 @@ it('should remember location between page visits', () => {
     .contains('Chapter 10. A Bosom Friend.')
     .should('be.visible');
 
-  cy.log('Go back home and to another book and then back');
-  cy.visit('/');
-  cy.findByRole('link', { name: 'Moby Dick' }).click();
```
Toxiapo: I don't think can navigate to a different page with mouse clicks without cypress complaining, I updated this code to visit the page directly.

kristojorg: I think we can, this seemed to work for me, no?

kristojorg: It was working locally at least, and it should work

kristojorg: But oh well this is fine too

Toxiapo: It was throwing [this](https://dashboard.cypress.io/projects/79fvf2/runs/318/test-results/c3741327-e33f-4eec-8faf-5ed67c8cf20f) error when running on Github actin. localhost seems fine.

kristojorg: Ah okay yea I _believe_ that is because it wasn't waiting long enough before trying to access the iframe. But I could be wrong. Anyways, this should be fine

PR #127, opened 2022-04-01
OE-340: use identifier as localstorage key if available

kristojorg:
This is a follow up to my save reading position that will use the `metadata.identifier` instead of the webpub manifest url if it is available as the local storage key.


kristojorg:
@Toxiapo idk why we started getting npm install dependency conflicts, but I updates some packages and hopefully that will fix it

Toxiapo (approved):
LGTM!

PR #128, opened 2022-04-06
Fix/location sync

kristojorg:
This allows disabling location and settings persistence, and exports functions to clear them. I changed to store all locations in a single object in local storage in order to make it easy to clear them.


Toxiapo (commented):
This looks great, just one question:

Does it make sense to persist the location query in the URL if `persistLastLocation` was disabled? Consider that the location query will always dictate what pages to start rendering.

Toxiapo (approved):

PR #129, opened 2022-04-26
OE-362: Detect book layout and apply CSS accordingly

Toxiapo:
This PR updates the code to detect the book layout and add injectables based on the book type presented on the webpub manifest. 

I am adding both `injectablesReflowable` and `injectablesFixed` as optional props because we want to allow consumer apps to add any injectables they wish to (assuming a host app wouldn't necessarily know if a book is fxl or reflowable) , but the reader should be responsible to determine which CSS to inject with based on the webpub manifest. 

Related: [PR ](https://github.com/NYPL/ePub-to-webpub/pull/21) exposes the layout property


src/HtmlReader/effects.ts
```diff
@@ -94,7 +94,79 @@ export function setCss(html: HTMLElement, settings: ReaderSettings): void {
...
+  // GLOBAL
   setCSSProperty(html, 'overflow', settings.isScrolling ? 'scroll' : 'hidden');
   // set the number of columns to only ever have 1.
   setCSSProperty(html, '--USER__colCount', '1');
 }
+
+/**
+ * Apply the transform property to the iframe document to fit the current screen viewport.
+ */
+export function setFXLCss(
```
Toxiapo: It might make more sense to call this function in the setCss function actually and detect the book layout and apply the CSS to the page, what do you think? @kristojorg 

Then the setCss function will accept a few more parameters (layout, viewport element)

Toxiapo:
@kristojorg Ideally, we want to calculate the scale property in a few places.

1. Before the contents get rendered on the screen from the resource fetching step. (useResource hook)
2. During window resize. Handle this in the useWindowResize hook.
3. _Perhaps_ when users change reading mode. It might make sense to have the page renders in full size on scrolling mode, but this is difficult, and I am not sure what the best way to handle this. One issue I can see from this is that when the user's on scrolling mode,  scrolled down halfway, and switched back to paginated, the page might end up showing only half the image.

src/HtmlReader/effects.ts
```diff
@@ -94,7 +94,79 @@ export function setCss(html: HTMLElement, settings: ReaderSettings): void {
...
+ * Extract the publication's width and height from the iframe viewport meta tag.
+ */
+function extractContentViewportSize(
+  iframeDocument: Document
+): { contentWidth: number | undefined; contentHeight: number | undefined } {
+  const viewport = iframeDocument?.querySelector('meta[name="viewport"]');
+  const content = viewport?.getAttribute('content');
+
+  const width = content?.match(/width=(\d+)/);
+  const height = content?.match(/height=(\d+)/);
```
Toxiapo, line 178: We trust that the publication will always include this viewport in their HTML files. Though I am unsure if they made this a standard in Epub doc. This only applies to epub2 and I can't find any resources mentiong this.

Toxiapo: Looks like this is what Readium mobile did as well. 
https://github.com/readium/swift-toolkit/blob/b4a9358ad9d0a7cb0072f5c9b1bbb72a7a2d28c7/Sources/Navigator/EPUB/Scripts/src/index-fixed-wrapper.js#L26

kristojorg (commented):
I have a question here. Right now, we call a `setFXLCss` effect in multiple places. Namely, when we get the resource initially, and when the viewport size changes. I'm wondering though, if instead we can simply use the normal react data flow to accomplish this? Can we calculate an `iframeScale` value and set that in the state object, and then simply pass it to the iframe's `style` prop? That way we just have to update the state value and don't have to think about manually updating the css property

kristojorg:
> Perhaps when users change reading mode. It might make sense to have the page renders in full size on scrolling mode, but this is difficult, and I am not sure what the best way to handle this. One issue I can see from this is that when the user's on scrolling mode, scrolled down halfway, and switched back to paginated, the page might end up showing only half the image.

I think FXL publications are assumed not to have. a scrolling/paginated mode. Our responsibility with these books is to show the entirety of the resource in the viewport, so there shouldn't be any scrolling (or other pages in one resource). The next page is the next resource

Toxiapo:
> I have a question here. Right now, we call a `setFXLCss` effect in multiple places. Namely, when we get the resource initially, and when the viewport size changes. I'm wondering though, if instead we can simply use the normal react data flow to accomplish this? Can we calculate an `iframeScale` value and set that in the state object, and then simply pass it to the iframe's `style` prop? That way we just have to update the state value and don't have to think about manually updating the css property

I have thought about this but I am not sure if we want to pass down the state this way.
We would have to calculate the initial scale when we get the resource initially to avoid the `flash` between the rendering of the content, and when the window resizes. If we do choose to store this state somewhere, where do we use it?

kristojorg (approved):
Ok this looks good to me! We just needs to bump the version by a major version due to the API change (`injectables` to `injectablesReflowable`) and alert DRB team what they should do to update. Also we should put instructions in the release when we do release it.

Toxiapo:
Not sure why the PDF tests are failing, but this update shouldn't affect PDF and it seems to be working locally. I will create a ticket to investigate. 

https://jira.nypl.org/browse/OE-498

PR #131, opened 2022-06-06
SFR-1412 Skip interstitial pages for ProjectMuse PDFs

jackiequach:
Handles start query parameter for manifest urls. Skips interstitial pages when using Next/Previous buttons and Table of Contents navigation.

### DIFF
--- a/src/PdfReader/index.tsx
+++ b/src/PdfReader/index.tsx
@@ -420,12 +420,10 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
     // do nothing if the reader is inactive
     if (state.state !== 'ACTIVE') return;
 
-    let startPage = 1;
-    if (manifest?.readingOrder && manifest?.readingOrder[state.resourceIndex]) {
-      startPage = getStartPage(
-        manifest?.readingOrder[state.resourceIndex].href
-      );
-    }
+    const startPage =
+      manifest?.readingOrder && manifest?.readingOrder[state.resourceIndex]
+        ? getStartPage(manifest?.readingOrder[state.resourceIndex].href)
+        : 1;
 
     if (state.pageNumber > startPage && !state.settings.isScrolling) {
       dispatch({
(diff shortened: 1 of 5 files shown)


src/PdfReader/index.tsx
```diff
@@ -406,7 +420,14 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
     // do nothing if the reader is inactive
     if (state.state !== 'ACTIVE') return;
 
-    if (state.pageNumber > 1 && !state.settings.isScrolling) {
+    let startPage = 1;
+    if (manifest?.readingOrder && manifest?.readingOrder[state.resourceIndex]) {
```
hokei: this is a preference, but I prefer 
```
const startPage = manifest?.readingOrder && manifest?.readingOrder[state.resourceIndex] ? getStartPage(
        manifest?.readingOrder[state.resourceIndex].href
      ); : 1
```

as this makes the startPage variable immutable.  
      
If it looks messy, I'd instead assign `manifest?.readingOrder[state.resourceIndex]` to a const and do the conditional on that.

src/PdfReader/index.tsx
```diff
@@ -459,17 +480,28 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
         const index = manifest?.readingOrder?.findIndex((link) => {
           return link.href === href;
         });
-        if (!index) {
+        if (index === undefined || index === -1) {
```
hokei: huh, this is tricky.  This will return "false" if index is `null` so it's not quite there.  I _think_ 
`if !(Number(index)) > 0) // or -1, the fact that pages that are 1-indexed is messing me up`  might give you what you want. 
Screen Shot 2022-06-07 at 12 45 25 PM

hokei:
Is there a place to add automated testing for this?

src/PdfReader/index.tsx
jackiequach: `!(Number(0) > 0)` returns true, so this doesn't work

hokei: That's an indexing problem, right?  If 0 is valid, then it would be `!(Number(index) >= 0)`

hokei: Since PDFs start at page=1 it messes with me every time I look at this repo

hokei:
The "Multi-page PDF" example in the test app is also a Muse PDF.  Is it possible to replace this manifest with a manifest that shows off this functionality?

hokei (approved):

PR #133, opened 2022-08-08
Add script injectables support

Toxiapo:
This PR updates the logic to support Javascript injectables.
i.e.
```ts
const injectables = [
  {
    type: "script", // make sure the type is correct
    url: `${origin}/js/sample.js`
  }
];
```

Motive: We need to inject javascript codes on CPW so that we can disable the copy&paste feature on the reader.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -67,7 +67,7 @@ jobs:
         run: npm i --omit=optional
 
       - name: Run Chrome
-        uses: cypress-io/github-action@v2
+        uses: cypress-io/github-action@v4.0.1
         env:
           CYPRESS_BASE_URL: ${{ github.event.deployment_status.target_url }}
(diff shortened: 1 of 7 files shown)


.github/workflows/cypress.yml
```diff
@@ -77,9 +77,10 @@ jobs:
         with:
           install: false
           browser: chrome
+          headed: true
```
Toxiapo, line 80: Unfortunately, running cypress in headless mode causes the bug where the single PDF won't be able to get the content from an external API, so I had to run this in headed mode. I think this is fine as far as regression, but please let me know if there are reasons we should run them in headless!

samanthaandrews: I think headed is fine!

samanthaandrews (approved):
LGTM

PR #134, opened 2022-08-10
Update Skeleton components so that it respect the colorMode setting

Toxiapo:
This PR updates the skeleton component to use the current colorMode setting instead of always showing white color background. Also refactored the Footer section into it's own component


samanthaandrews (approved):
🚀

PR #135, opened 2022-08-15
Refactor Iframe link click

Toxiapo:
This PR fixes the bug where sometimes when the user clicks on an anchor tag from the book content, it throws the "invalid URL" error. This is because some of the `<a>` tags in the book do not come with a href value, therefore, when constructing a URL with the href value via `new URL(href)`,  we get a type error.

There are a couple of changes in this PR:

- We should check to ensure the `<a>` tag contains a href value in our handleLinkClick function. `<a>` without the href attribute is semantically correct and valid according to the [HTML standard](https://www.w3.org/TR/2016/REC-html51-20161101/textlevel-semantics.html#the-a-element). So we will just let it pass through if it's missing the href. Otherwise, we perform our normal link click action.
- Changing the code to use `evt.currentTarget` to get the href value in our event listener, because sometimes there are nested elements inside the anchor element.
- While fixing the bug, I realized that our app only recognizes links that are from the book(TOC), while others, like https://google.com was ignored. I added a function to detect if the clicked link is a valid resource from the book; if not, we just open the link in a separate tab.


samanthaandrews (approved):
LGTM

PR #137, opened 2022-10-07
Remove Service Workers

Toxiapo:
Ticket: [OE-272](https://jira.nypl.org/browse/OE-272)

This PR removes the service worker and workbox from the repo.


example/axisnow-encrypted.tsx
```diff
@@ -73,7 +73,7 @@ const AxisNowEncrypted: React.FC<AxisNowEncryptedProps> = (props) => {
 
   return (
     <WebReader
-      injectables={props.injectables}
+      injectablesReflowable={props.injectablesReflowable}
```
kristojorg: This is a breaking change right? The old `injectables` prop no longer works correct? If so, then we will need to bump the version number, provide a little migration guide (just a list of steps) which we can provide to the DRB team

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nypl/web-reader",
-  "version": "3.2.3",
+  "version": "3.3.0",
```
kristojorg: Will likely need to go to v4

package.json
```diff
@@ -84,6 +84,7 @@
     "@babel/core": "^7.12.10",
     "@emotion/jest": "^11.8.0",
     "@parcel/transformer-typescript-tsc": "^2.4.1",
+    "@parcel/transformer-typescript-types": "^2.4.1",
```
kristojorg: Do we need both of these?

kristojorg (commented):
Okay this looks good, just the questions about the `injectablesReflowable` change. Did we change that API previously or does both `injectables` and `injectablesReflowable` still work? I don't quite remember

example/axisnow-encrypted.tsx
Toxiapo: yup, the `injectables` no longer work, but we already bumped the version to 3 for that upgrade.

kristojorg: Oh I see okay

package.json
Toxiapo: I think it was added by mistaken. Deleting now!

kristojorg (approved):
Looks good to me! When we publish this, let's make sure the change is recorded in the Github Release and ping the DRB team to let them know (though I don't think they used this functionality).

kristojorg (approved):

kristojorg (approved):

PR #140, opened 2022-11-28
OE-685: update settings panel

oliviawongnyc:
This PR contains the work for ticket [OE-685](https://jira.nypl.org/browse/OE-685), which is implementing new designs for the web reader settings panel. I do not have designs for the PdfSettings, so this change is mainly for the HtmlSettings. This is a big PR, and my first for the web reader, so appreciate your feedback; please let me know if you have any questions!


kristojorg:
One thing I noticed here is that I think the popover is overflowing the page: 

image

This is causing the gap to the right. It's also there when the modal isn't open: 

image

You may need to change how the popover anchors to the button. I think there should be a way to make the right side of the modal align with the button instead of the left

oliviawongnyc:
> One thing I noticed here is that I think the popover is overflowing the page:
> 
> image
> 
> This is causing the gap to the right. It's also there when the modal isn't open:
> 
> image
> 
> You may need to change how the popover anchors to the button. I think there should be a way to make the right side of the modal align with the button instead of the left

@kristojorg, I think it looks different on Chrome and Safari:

Screen Shot 2022-11-30 at 4 25 08 PM

Screen Shot 2022-11-30 at 4 24 53 PM

src/HtmlReader/reducer.ts
```diff
@@ -482,6 +482,18 @@ export default function makeHtmlReducer(
...
 
+      case 'RESET_FONT_SIZE':
+        if (state.state === 'INACTIVE') {
+          return handleInvalidTransition(state, action);
+        }
+        return {
+          ...state,
+          settings: {
+            ...state.settings,
+            fontSize: 100,
```
Toxiapo: We should get this default value from the "constant.ts" file

src/ui/HtmlSettings.tsx
```diff
@@ -1,111 +1,237 @@
...
 export type HtmlSettingsProps = {
+  buttonTextColor: string;
+  checkedButtonBgColor: string;
   navigator: HtmlNavigator;
+  iconFill: string;
   readerState: ReaderState;
   paginationValue: string;
 };
 
+const fontInfo = {
```
Toxiapo: I think we can move this to the `constant.ts` file

src/ui/HtmlSettings.tsx
```diff
@@ -1,111 +1,237 @@
...
+    fontSize !== 100 ||
+    fontFamily !== 'publisher' ||
+    isScrolling;
+
+  const resetSettings = () => {
+    setColorMode('day');
+    setFontFamily('publisher');
+    setScroll('paginated');
+    resetFontSize();
+  };
```
Toxiapo: Should the button reset everything or just the font size?

If it resets all the settings, we should change the reducer from `RESET_FONT_SIZE` to `RESET_SETTINGS` and just reset the state there, so we don't need to create this function here.

src/ui/HtmlSettings.tsx
```diff
@@ -1,111 +1,237 @@
...
     <>
       <ToggleGroup
         value={fontFamily}
         label="text font options"
         onChange={setFontFamily}
       >
-        <ToggleButton value="publisher" label="Publisher">
-          Publisher
+        <ToggleButton value="publisher" label="Default" fontSize={[-1, -1, 0]}>
+          <Text>Default</Text>
```
Toxiapo: We can also add the `Text` inside the ToggleButton component since it's now used for all ToggleButtons

src/ui/SettingsButton.tsx
```diff
@@ -8,12 +8,13 @@ import {
...
 } from '@chakra-ui/react';
 import { PDFActiveReader, HTMLActiveReader } from '../types';
-import { MdOutlineSettings, MdOutlineCancel } from 'react-icons/md';
 
 import Button from './Button';
 import useColorModeValue from './hooks/useColorModeValue';
 import PdfSettings from './PdfSettings';
 import HtmlSettings from './HtmlSettings';
+import { ReaderSettings } from './icons';
+import Header from './Header';
```
Toxiapo: lint: Header is never used

package.json
```diff
@@ -145,7 +145,7 @@
     "ts-jest": "^27.1.4",
     "ts-node": "^10.1.0",
     "tslib": "^2.3.1",
-    "typescript": "^4.3.5"
+    "typescript": "4.4.3"
```
Toxiapo, line 148: I am curious what's the reason for the fixed version

README.md
```diff
@@ -123,7 +123,7 @@ Finally, to use in a vanilla Javascript app:
 Most styling is included in the basic UI, but we also ship a few css files that must be included:
 
 1. Both the HTML and the PDF side have css that is necessary to be included for the dependencies we use to render correctly. This is built automatically into `@nypl/web-reader/dist/esm/index.css` and `@nypl/web-reader/dist/cjs/index.css`. Depending which package you are using, you should include one of those files in your bundle.
-1. The HTML Reader can inject `<style>` tags (and other tags) into the reader iframe itself, called an "injectable". This is used to add styles to the html content of the publication. More on this is below.
+2. The HTML Reader can inject `<style>` tags (and other tags) into the reader iframe itself, called an "injectable". This is used to add styles to the html content of the publication. More on this is below.
```
Toxiapo: Just read about [this](https://github.com/NYPL-Simplified/circulation-patron-web/pull/532#discussion_r1039494992) on @kristojorg comment, I think it is a nice trick : )

src/ui/HtmlSettings.tsx
Toxiapo: Ok Just confirmed with Miguel that the reset button should reset everything. We should be able to just set the default setting states inside this new reducer.

Toxiapo (changes requested):
Just a couple of comments and questions:

1. Can you make the font-family buttons smaller and the rest of the buttons a little bigger on the web version? 
2. On sepia/night mode, the popover background colors should also be updated accordingly. 
3. Can you add an instruction/example to show how the consuming app can use the open dyslexic font in their application? (e.g. for nextjs app, you should have the `fonts` folder in your /public folder that includes the fonts found in `@nypl/web-reader/xxx`)

Otherwise this is looking great!

cypress/integration/html-reader/settings.ts
```diff
@@ -95,7 +95,7 @@ it('should maintain settings between page visits', () => {
   cy.log('Settings overlay should show correct values');
   cy.findByRole('button', { name: 'Settings' }).click();
   cy.wait(100);
-  cy.findByRole('radio', { name: 'Publisher' }).should('be.checked');
+  cy.findByRole('radio', { name: 'Default' }).should('be.checked');
   cy.findByRole('radio', { name: 'Day' }).should('be.checked');
   cy.findByRole('radio', { name: 'Paginated' }).should('be.checked');
```
Toxiapo, line 122: Let's also add a test for the reset button

src/ui/Header.tsx
```diff
@@ -58,17 +78,22 @@ export default function Header(
         />
         <SettingsCard {...props} />
         <Button
+          aria-expanded={isFullscreen}
+          aria-label="Toggle full screen"
```
Toxiapo, line 82: nice!

README.md
oliviawongnyc: oooooh.

package.json
oliviawongnyc: Sam and I made this decision together. I had npm installed again when I was trying to npm link and lots of things went wrong. I noticed that the typescript version in package-lock had increased so we pinned the old one and everything worked again.

oliviawongnyc:
> Just a couple of comments and questions:
> 
> 1. Can you make the font-family buttons smaller and the rest of the buttons a little bigger on the web version?
> 2. On sepia/night mode, the popover background colors should also be updated accordingly.
> 3. Can you add an instruction/example to show how the consuming app can use the open dyslexic font in their application? (e.g. for nextjs app, you should have the `fonts` folder in your /public folder that includes the fonts found in `@nypl/web-reader/xxx`)
> 
> Otherwise this is looking great!

@Toxiapo, could you please elaborate on the first point? Thank you!

Toxiapo:
@oliviawongnyc yea, I am just noticing that the first row has smaller paddings than the rest in the design

Screen Shot 2022-12-08 at 2 56 32 PM

src/ui/SettingsButton.tsx
```diff
@@ -26,42 +26,50 @@ export default function SettingsCard(
...
   return (
     <>
       <Popover
-        placement="bottom"
+        gutter={0}
+        placement="bottom-end"
         isOpen={isOpen}
         onOpen={open}
         onClose={close}
+        offset={[-200, 0]}
```
Toxiapo: This fixes the overflow issue though I think this is a bug from chakraui that we should probably add a ticket to address.  @kristojorg, what do you think? It might be tricky because it's browser specific, firefox and chrome seems fine.

Toxiapo: @oliviawongnyc Let's try one more thing. Remove this offset and give the `PopoverContent`  a `width="inherit"` props. I think this fixes the overflow bug.

It seems like it could be an issue with the width: https://github.com/chakra-ui/chakra-ui/issues/4104#issuecomment-853764916

Toxiapo (commented):
Just that one last comment then we can pass this to QA!

README.md
```diff
@@ -125,6 +125,8 @@ Most styling is included in the basic UI, but we also ship a few css files that
 1. Both the HTML and the PDF side have css that is necessary to be included for the dependencies we use to render correctly. This is built automatically into `@nypl/web-reader/dist/esm/index.css` and `@nypl/web-reader/dist/cjs/index.css`. Depending which package you are using, you should include one of those files in your bundle.
 1. The HTML Reader can inject `<style>` tags (and other tags) into the reader iframe itself, called an "injectable". This is used to add styles to the html content of the publication. More on this is below.
 
+In order for the OpenDyslexic font to be displyed in the Settings panel as intended, you must include the `fonts/opendyslexic` folder and its contents in your entry directory (in a Next.js app, this would be in your `/public` folder). You can find this folder in `@nypl/web-reader/static`.
```
kristojorg: I think this is true of all the fonts (except publisher) actually, not just open dyslexic.

src/ui/Header.tsx
```diff
@@ -58,17 +78,22 @@ export default function Header(
...
+              as={isFullscreen ? ToggleFullScreenExit : ToggleFullScreen}
+              fill={iconFill}
               w={6}
               h={6}
             />
           }
         >
-          <Text variant="headerNav">Toggle Fullscreen</Text>
+          <Text variant="headerNav">
+            {isFullscreen ? 'Full screen exit' : 'Full screen'}
```
kristojorg, line 95: Should this say "Exit full screen"?

src/ui/HtmlSettings.tsx
```diff
@@ -1,112 +1,192 @@
...
+  Reset,
+  Sepia,
+} from './icons';
+import { FONT_DETAILS } from '../constants';
 import ToggleButton from './ToggleButton';
 import ToggleGroup from './ToggleGroup';
 
 export type HtmlSettingsProps = {
+  buttonTextColor: string;
+  checkedButtonBgColor: string;
```
kristojorg: Just curious why we pass these in? A comment would be useful here

src/ui/HtmlSettings.tsx
```diff
@@ -1,112 +1,192 @@
...
+    resetSettings,
     setColorMode,
     setScroll,
   } = navigator;
 
+  const settingsHaveChanged =
+    colorMode !== 'day' ||
+    fontSize !== 100 ||
+    fontFamily !== 'publisher' ||
+    isScrolling;
```
kristojorg: If this is meant to check that the current settings are not the default settings, let's compare with the `DEFAULT_SETTINGS` constant, and maybe extract this into a function `isDefaultSettings` to make it more clear

src/ui/HtmlSettings.tsx
```diff
@@ -1,112 +1,192 @@
...
-        <ToggleButton value="sans-serif" label="Sans-Serif">
-          Sans-Serif
-        </ToggleButton>
-        <ToggleButton value="open-dyslexic" label="Dyslexia-Friendly">
-          Dyslexia-Friendly
-        </ToggleButton>
+        <ToggleButton
+          value="publisher"
+          label="Default"
+          fontSize={[-1, -1, 0]}
```
kristojorg: Just an FYI, I think you could apply this `fontSize` to the `ToggleGroup` component and it _should_ then apply to all the children (unless they have more specific css setting their font size). In a situation like this I would try to find a way to not repeat this fontsize in multiple places

src/ui/HtmlSettings.tsx
```diff
@@ -1,112 +1,192 @@
...
+        />
+        <ToggleButton
+          value="serif"
+          label="Serif"
+          font="georgia"
```
kristojorg: Should the font here be the `serif` font from the theme (which should be georgia I believe)? Would rather extract from the theme than hardcode these so that if we choose to change our serif font to something different, we simply change it in the theme once

src/ui/HtmlSettings.tsx
```diff
@@ -1,112 +1,192 @@
...
-        <Button
-          flexGrow={1}
-          aria-label="Increase font size"
-          onClick={increaseFontSize}
-          variant="toggle"
+          {FONT_DETAILS[fontFamily].heading}
+        </Heading>
+        <Text
+          color={buttonTextColor}
+          fontFamily={FONT_DETAILS[fontFamily].token}
```
kristojorg, line 93: These could be defined in the theme as a [text style](https://chakra-ui.com/docs/styled-system/text-and-layer-styles#text-styles) which would then allow you to just pass a single prop to the component `textStyle="serif"` and it would apply all of the serif styles (font weight, family, etc). Just an idea

src/ui/HtmlSettings.tsx
kristojorg: I realized its so we only calaculate them in one place (ui/SettingsButton) and use them in HtmlSettings and PdfSettings. I think a comment to that effect would be useful

src/ui/SettingsButton.tsx
kristojorg: Ok I think I found the solution here and make a PR for it

src/ui/ToggleButton.tsx
```diff
@@ -1,20 +1,27 @@
...
 import { ColorMode } from '../types';
-import { MdOutlineCheckCircle } from 'react-icons/md';
 
 export interface ToggleButtonProps
   extends React.ComponentPropsWithoutRef<typeof ChakraBox> {
-  isChecked?: false;
+  bgColor?: string;
+  font?: string;
+  fontSize?: number;
+  fontWeight?: string;
```
kristojorg: Are these not already props on the `ChakraBox`? Do we need to explicitly add them?

src/ui/ToggleButton.tsx
```diff
@@ -23,33 +30,53 @@ export interface ToggleButtonProps
...
+    ...rest
+  } = props;
   const { getInputProps, getCheckboxProps } = useRadio(props);
 
   const input = getInputProps();
   const checkbox = getCheckboxProps();
   const theme = useTheme();
+  const buttonStyles = useStyleConfig('Button', {
+    variant: 'settings',
+    font: font || 'roboto',
```
kristojorg: I don't think we should be hard coding a font in here. Is the intention that if it is not set directly then we use the default? If so then I think we should use `initial` maybe

src/ui/ToggleButton.tsx
```diff
@@ -23,33 +30,53 @@ export interface ToggleButtonProps
...
+    fontWeight: fontWeight || 'light',
+    permanentBgColor: bgColor || undefined,
+    permanentTextColor: textColor || undefined,
+    value,
+  });
 
   return (
     // This will override the default theme if we specify the colorMode to the toggle button.
     <ThemeProvider theme={getTheme(colorMode ?? theme.currentColorMode)}>
+      <Fonts />
```
kristojorg, line 38: Should this be rendered higher up in the tree? Maybe even at the root since it is not just the toggle button that needs these fonts right?

src/ui/ToggleButton.tsx
```diff
@@ -23,33 +30,53 @@ export interface ToggleButtonProps
...
-              position="absolute"
+              as={icon}
               verticalAlign="middle"
-              right={2}
-              top="50%"
-              transform="translateY(-50%)"
-              alignItems="baseline"
-              w={5}
-              h={5}
+              mr="6px"
```
kristojorg: This is a theme value of `1.5`

src/ui/nypl-base-theme/foundations/typography.ts
```diff
@@ -13,6 +13,10 @@ const typography: Typography = {
     body: '-apple-system, BlinkMacSystemFont, system-ui, sans-serif',
     heading: '-apple-system, BlinkMacSystemFont, system-ui, sans-seri',
     mono: `Courier,SFMono-Regular,Menlo,Monaco,Consolas,"Liberation Mono","Courier New",monospace`,
+    georgia: `'Georgia', serif`,
+    helvetica: `'Helvetica', 'Arial', serif`,
+    opendyslexic: `'OpenDyslexic', sans-serif`,
+    roboto: `'Roboto', sans-serif`,
```
kristojorg: I think we should make changes to our `theme` instead of the `nypl-base-theme`. That one is copied over from the design system and the idea was just to extend from it (not modify it) in the web reader. Eventually we should just import that base theme instead of using a copied over version

src/ui/theme/components/button.ts
```diff
@@ -30,28 +30,43 @@ const getButtonStyle = (getColor: GetColor) =>
...
  ** Hovered
  */
+
 const variantSolid = (getColor: GetColor) => (
   props: React.ComponentProps<typeof Button>
 ) => {
-  const bgColor = getColor('ui.white', 'ui.black', 'ui.sepia');
+  const { 'aria-expanded': expanded, isSettingsButton } = props;
+  const bgColor = getColor(
+    isSettingsButton ? 'ui.white' : 'ui.gray.light-warm',
```
kristojorg: If it is a settings button would it not use the `variantSettings`? Wondering if this check is necessary basically

src/ui/theme/components/button.ts
```diff
@@ -61,30 +76,75 @@ const variantSolid = (getColor: GetColor) => (
...
 
-const variantToggle = (getColor: GetColor) => (
+const variantSettings = (getColor: GetColor) => (
   props: React.ComponentProps<typeof Button>
 ) => {
-  const bgColor = getColor('ui.white', 'gray.800', 'ui.white');
-  const color = getColor('gray.800', 'ui.white', 'gray.800');
+  const {
+    permanentBgColor,
+    permanentTextColor,
```
kristojorg: A comment about what permanent means here would be helpful : )

kristojorg: Or wherever you define the type for this prop would also work

src/ui/theme/components/button.ts
```diff
@@ -61,30 +76,75 @@ const variantSolid = (getColor: GetColor) => (
...
+    'ui.gray.light-warm',
+    'ui.gray.x-dark',
+    'ui.sepiaChecked'
+  );
+
+  const isFontToggle =
+    value === 'publisher' ||
+    value === 'serif' ||
+    value === 'sans-serif' ||
+    value === 'open-dyslexic';
```
kristojorg: I don't think this logic should live in the theme. Instead, you can pass the styles you need to the font toggle when you are rendering it `<Toggle py={6}>`. The theme shouldn't be making decisions on the component based on what it infers the type is from the value. The idea behind the theme is to create components that are not aware of the specific ways they are used but just set general styles. Not sure if I'm explaining this well, but I this well but I think all the value checks can be replaced by simply passing custom styles to the component when you render it (and know exactly what type of button it is)

src/ui/theme/components/button.ts
kristojorg: If the idea is to be able to pass a `permanentBgColor` that overrides the color set by the colorMode, I think you can do that by passing a `bgColor` at render time: `<ToggleButton bgColor="ui.white">` should render a ToggleButton with white background no matter what the `colorMode` is at a given time.

kristojorg (changes requested):
This is fantastic work! The settings panel and the whole chrome looks much better. Thanks for all your patience and work on this : )

I have a few comments and changes I'd like to make before merging this. I also made a PR to your branch updating the `Popover` with a fix I discovered.

If you want to discuss any of the comments, please reach out and let's set up a time to do that. I'm always happy to chat about this stuff and maybe my explanations will make more sense in a meeting than over GH

src/ui/HtmlSettings.tsx
kristojorg: Of if a consuming app wants to change their serif font they can simply provide an updated theme

src/ui/Header.tsx
oliviawongnyc: Not according to this design: https://www.figma.com/file/VhO0lUqJbFut5sJGuThBvb/Web-Reader-settings?node-id=1659%3A206&t=RBvpV07tT7wO7zPm-0

src/ui/SettingsButton.tsx
oliviawongnyc: Thank you, both! I merged in your PR @kristojorg. Looks great.

README.md
oliviawongnyc: I think I'm a little confused about the fonts. I'll put some time on for us to discuss this IRL!

src/ui/HtmlSettings.tsx
oliviawongnyc: I'm actually not using these values in PdfSettings. I couldn't use the useColorModeValue hook in the HtmlSettings component because it is called conditionally. We can chat tomorrow about whether you would accomplish this differently.

README.md
oliviawongnyc: Expanded this comment per our conversation.

src/ui/HtmlSettings.tsx
oliviawongnyc: Moved this calculation to the HtmlSettings component per our conversation.

src/ui/nypl-base-theme/foundations/typography.ts
oliviawongnyc: Okay! I'll move over the colors I added to nypl-base-theme, as well.

src/ui/theme/components/button.ts
oliviawongnyc: This is a remnant of an earlier styling attempt and is no longer needed. I deleted!

src/ui/theme/components/button.ts
oliviawongnyc: I understand! We can discuss what I did to change this at our meeting.

src/constants.ts
```diff
@@ -12,7 +12,7 @@ export const CHROME_HEIGHT = HEADER_HEIGHT + FOOTER_HEIGHT;
 export const DEFAULT_HEIGHT = `calc(100vh - ${CHROME_HEIGHT}px)`;
 export const DEFAULT_SHOULD_GROW_WHEN_SCROLLING = true;
 
-export const DEFAULT_SETTINGS: ReaderSettings = {
```
kristojorg: Reminder to put the `ReaderSettings` type back onto this after we took it off for our testing : )

src/ui/SettingsButton.tsx
```diff
@@ -73,6 +78,7 @@ export default function SettingsCard(
             {props.type === 'HTML' && (
               <HtmlSettings
                 navigator={props.navigator}
+                iconFill={iconFill}
```
kristojorg, line 81: I think calculation of this can also be moved inside of `<HtmlSettings>` as well

src/ui/nypl-base-theme/foundations/typography.ts
```diff
@@ -4,16 +4,16 @@ type Typography = {
...
-  fonts: {
-    body: '-apple-system, BlinkMacSystemFont, system-ui, sans-serif',
-    heading: '-apple-system, BlinkMacSystemFont, system-ui, sans-seri',
-    mono: `Courier,SFMono-Regular,Menlo,Monaco,Consolas,"Liberation Mono","Courier New",monospace`,
-  },
+  // fonts: {
+  //   body: '-apple-system, BlinkMacSystemFont, system-ui, sans-serif',
+  //   heading: '-apple-system, BlinkMacSystemFont, system-ui, sans-seri',
+  //   mono: `Courier,SFMono-Regular,Menlo,Monaco,Consolas,"Liberation Mono","Courier New",monospace`,
+  // },
```
kristojorg: Did you mean to comment these out?

src/constants.ts
oliviawongnyc: Yes! Reverted back. Thank you.

src/ui/theme/components/button.ts
```diff
@@ -61,30 +74,52 @@ const variantSolid = (getColor: GetColor) => (
...
-    bgColor,
+    bgColor: getColor('ui.white', 'ui.black', 'ui.sepia'),
+    border: '1px solid',
+    borderBottom: (value === 'paginated' || value === 'scrolling') && 'none',
+    borderRadius:
+      value === 'paginated'
+        ? '0 0 0 4px'
+        : value === 'scrolling'
+        ? '0 0 4px 0'
+        : null,
```
kristojorg: A couple more places where we are checking the value that I'd like to remove and instead have at the point the button is rendered

src/ui/theme/components/button.ts
```diff
@@ -61,30 +74,52 @@ const variantSolid = (getColor: GetColor) => (
...
       bgColor,
     },
-    _active: {
+    _focus: {
       bgColor,
-    },
-    _checked: {
-      color: 'ui.white',
-      bgColor: 'green.700',
+      zIndex: 1000,
```
kristojorg: Out of curiosity, what was it that required us to set a z index like this when the button is focused? I think this button shouldn't have a z index at all preferably. It should inherit that if it's parent. That way if it is rendered eg in the header, it will have the z index of the header and if it's rendered in a modal it will have the z index of the modal. If we set it explicitly then consider the situation where we render it in the body of the page and then open a modal on top of the page. The modal might have z index of 500 whereas this button has 1000 and therefore is going to float above the modal. Does that make sense?

src/ui/theme/index.ts
```diff
@@ -20,7 +22,9 @@ import { Dict } from './types';
 export function getTheme(colorMode: ColorMode = 'day'): Dict<unknown> {
   return extendTheme(
     {
+      breakpoints,
```
kristojorg: We shouldn't need to place this here since we are already extending from the nypl base theme. So we shouldn't need to redefine the breakpoints from the base theme

kristojorg (commented):
Just a few comments : )

src/ui/SettingsButton.tsx
oliviawongnyc: Good call

oliviawongnyc: Oh wait, I'm actually using it in SettingsButton for the settings button icon.

src/ui/nypl-base-theme/foundations/typography.ts
oliviawongnyc: No, this was me testing. Thanks for catching.

src/ui/theme/components/button.ts
oliviawongnyc: Removed!

src/ui/SettingsButton.tsx
kristojorg: Ok sounds good!

src/ui/theme/components/button.ts
oliviawongnyc: This is the problem I was trying to solve: https://twitter.com/adamwathan/status/1232354577472684033?lang=en

When tabbing, the focus ring is wonky. It only shows on some of the sides of each button because of how the elements overlap. I'm trying to come up with other solutions. Let me know if knowing the issue now makes you think of something better!

Toxiapo (approved):
This is amazing! 🚀

kristojorg (approved):
So so good! Thanks for persisting thru the back and forth... it was worth it. This is great. Excited to have this in OEW : )

kristojorg:
Once we merge, we should release a new version. You can do that by running `npm run release` on the main branch

oliviawongnyc:
> So so good! Thanks for persisting thru the back and forth... it was worth it. This is great. Excited to have this in OEW : )

Thank YOU for all your great feedback! Proud of this!

oliviawongnyc:
> This is amazing! 🚀

Thank you for all your help, Jiayong!!

PR #142, opened 2022-12-08
PDF table of contents feature branch

jackiequach:
Feature branch for [SFR-1522](https://jira.nypl.org/browse/SFR-1522)

This PR includes changes from #150 and #151 (summarized):
- Adds the `addTocToManifest` helper function.
- Extracts the reducer, types, and utility functions into separate files for easier understanding.
- Makes a GO_FORWARD GO_BACKWARD and GO_TO_HREF action.
- Moves all action logic into the reducer (out of the callback) so it can be shared easier and centralized.
- Makes an effect that scrolls the user to the correct page when the page number changes.


kristojorg (approved):
Looks great!

PR #144, opened 2022-12-21
SFR-1524/1525: Implement PDF ToC component 

jackiequach:
[SFR-1524](https://jira.nypl.org/browse/SFR-1524) and [SFR-1525](https://jira.nypl.org/browse/SFR-1525)

This PR does the following:
- Adds functionality to get the sub-chapters to `generateSinglePdfToc`.
- Pulls the styles from `TableOfContent` into a theme file to be reused in `PdfTableOfContent`.
- Separates `MissingToc` and `Item` into individual files.
- Implements `PdfTableOfContent` in the `Header`.

### DIFF
--- a/src/PdfReader/generateSinglePdfToc.tsx
+++ b/src/PdfReader/generateSinglePdfToc.tsx
@@ -14,8 +14,8 @@ export default async function generateSinglePdfToc(
     const outline = await pdf.getOutline(); // get the TOC outline
 
     if (outline) {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       const addTocItem = async (
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         outline: any[],
         tocItems: PdfTocItem[]
       ): Promise<PdfTocItem[]> => {
--- a/src/PdfReader/generateSinglePdfToc.tsx
+++ b/src/PdfReader/generateSinglePdfToc.tsx
@@ -1,9 +1,5 @@
 import { pdfjs } from 'react-pdf';
-
-type PdfTocItem = {
-  title: string;
-  pageNumber: number;
-};
+import { PdfTocItem } from '../types';
 
 /**
  * Retrieve TOC data from single-resource PDF using PDF.js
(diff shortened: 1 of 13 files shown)


samanthaandrews (commented):
This looks good to me, but I think I will feel more confident approving it if we review it together on a quick call with me, you, @Toxiapo, and @kristojorg since I have less familiarity with the web-reader than they do. I'll set up time for next week when I'm back in the office on Thursday.

I did some manual testing and noticed that in scrolling mode, clicking on items in the TOC doesn't navigate users to the page. I was unsure if this is something that will be addressed in a future PR/ticket.

Toxiapo:
I agree with @samanthaandrews, I think a call is a good idea.

src/PdfReader/generateSinglePdfToc.tsx
```diff
@@ -13,31 +9,46 @@ type PdfTocItem = {
...
-        if (dest && dest.length > 0) {
-          // get the page referance
-          const ref = dest[0];
-          const pageIndex = await pdf.getPageIndex(ref);
-          if (pageIndex) {
-            tocItems.push({
-              title: outline[i].title,
-              pageNumber: pageIndex + 1,
-            });
+      const addTocItem = async (
```
kristojorg, line 17: Curious why this code is nested inside a function definition?

kristojorg (changes requested):
I agree that a call would be good. My main thought here is that I would like to keep the TOC implementation between formats as similar as possible. Our primary point of abstraction between formats is the `WebpubManifest`, which has the ability to hold a table of contents. So I'd like to try to fit the PDF TOC into this format so that it can be used by the app without needing to consider the special case of a single pdf toc everywhere. 

I'm wondering:

1. Could we implement this when generating the manifests initially? This would be the ideal scenario. There is currently a pdf-to-webpub script somewhere that generates `WebpubManifest` JSON files for the pdfs. We could add code there to generate the TOC and thus this complexity would be handled beforehand and not in the browser.
2. If not that, could we instead export a utility to add a TOC to the manifest? The consumer would then use it like this:
```ts
const manifestWithTOC = addTocToManifest(manifest, pdf);

// later
<WebReader
  manifest={manifestWithToc}
/>
```
This way the web reader code doesn't need to care about whether the TOC is a single-pdf-toc or a multi-pdf-toc or an epub-toc or what. It just sees the `toc` in the manifest and uses it. 

Let's schedule a time to chat and figure out the best approach : )

src/PdfReader/generateSinglePdfToc.tsx
jackiequach: There is no importable type for `PDFDocumentProxy`, so I nested the function to not need to pass `pdf` as type `any`.  I can update it to be a separate function definition and disable the type warning.

jackiequach:
Closing this PR as this implementation will not be used

PR #147, opened 2023-01-18
OE-685: Makes adjustments per VQA feedback

oliviawongnyc:
VQA feedback here: https://www.figma.com/file/VhO0lUqJbFut5sJGuThBvb/Web-Reader-settings?node-id=1823%3A808&t=pf7UQORauRlAyub9-0

PR #148, opened 2023-01-18
Adds workaround for Chakra/Safari bug

oliviawongnyc:
Added a workaround to fix Chakra bug where Settings button would open, but not close, the popover. 
Everything is still working as expected in Chrome with this change.

Issue is talked about in various places, including: https://github.com/chakra-ui/chakra-ui/issues/3461
Used a solution that works for the Menu component found here: https://github.com/chakra-ui/chakra-ui/issues/587

Resolves Jira ticket OE-774: https://jira.nypl.org/browse/OE-774

### DIFF
--- a/src/ui/SettingsButton.tsx
+++ b/src/ui/SettingsButton.tsx
@@ -40,10 +40,11 @@ export default function SettingsCard(
     <>
       <Popover
         gutter={0}
+        closeOnBlur
         placement="bottom-start"
         isOpen={isOpen}
-        onOpen={open}
         onClose={close}
+        onOpen={open}
         autoFocus={true}
         preventOverflow
         strategy="fixed"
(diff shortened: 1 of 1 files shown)


src/ui/SettingsButton.tsx
```diff
@@ -40,17 +40,21 @@ export default function SettingsCard(
...
         onClose={close}
+        onOpen={open}
         autoFocus={true}
         preventOverflow
         strategy="fixed"
       >
         <PopoverTrigger>
           <Button
             onClick={open}
+            onMouseDown={(e: React.MouseEvent<HTMLButtonElement>) =>
```
samanthaandrews, line 63: let's add a comment here describing why we've added this & link to the issues you found, so we don't forget why it's here

samanthaandrews (approved):
Just that one small thing, otherwise I think this can move to QA. Feel free to assign the Jira ticket to Cal & maybe Slack him to let him know you're doing so.

src/ui/SettingsButton.tsx
oliviawongnyc: Added!

kristojorg (approved):

kristojorg:
Looks great! Feel free to merge (squash and merge please). We can then wait to merge https://github.com/NYPL-Simplified/web-reader/pull/147 before releasing this to npm probably

PR #150, opened 2023-01-30
SFR-1597: Add TOC helper function and implementation

jackiequach:
Resolves Jira ticket [SFR-1597](https://jira.nypl.org/browse/SFR-1597)

Renamed `generateSinglePdfToc` to `addTocToManifest` and commented out sub-chapter handling due to how `goToPage` is implemented. `goToPage` will need to be updated to recursively search `manifest.toc` for a matching child ref.

Since the manifest is being parsed in web reader (`webpubManifestUrl` is passed to WebReader from consuming app), we are unable to modify the manifest prior to rendering the web reader without reworking the props passed to `WebReader`. Added `addTocToManifest` implementation to `usePdfReader`.

Updated `goToPage` to handle single PDF navigations without needing to reload the resource and parse the page hashes for update `state.page` and scroll position. 

Added integration tests for single pdf TOC.


src/PdfReader/index.tsx
```diff
@@ -472,36 +503,71 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
-        const index = manifest?.readingOrder?.findIndex((link) => {
-          return link.href === href;
-        }) as number;
-        if (index < 0) {
-          throw new Error('Cannot find resource in readingOrder');
+      if (isSinglePDF) {
+        const getIndexFromHref = (href: string): number => {
+          const index = manifest?.toc?.findIndex((link) => {
+            return link.href === href;
+          }) as number;
```
samanthaandrews: curious as to why we have to coerce `number` type here

samanthaandrews (commented):
This looks really good. I am going to let @kristojorg take a look before signing off.

src/PdfReader/index.tsx
jackiequach: `index` could be `undefined` if `manifest` or `manifest.toc` is undefined.

samanthaandrews: got it!

src/PdfReader/addTocToManifest.tsx
```diff
@@ -0,0 +1,64 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { pdfjs } from 'react-pdf';
+import { ReadiumLink } from '../WebpubManifestTypes/ReadiumLink';
+import { WebpubManifest } from '../types';
+
+const addTocItem = async (
+  pdf: any,
+  outline: any[],
```
kristojorg: Do we need to use `any` type here? I would rather explicitly type these if possible

src/PdfReader/addTocToManifest.tsx
```diff
@@ -0,0 +1,64 @@
...
+      }
+    }
+  }
+  return Promise.all(tocItems);
+};
+
+/**
+ * Adds TOC data to Webpub Manifest from single-resource PDF using PDF.js
+ * @param manifest
+ * @param proxiedUrl
```
kristojorg: Can you explain what the `proxiedUrl` is in this comment? What is it a url to and how/why is it proxied?

src/PdfReader/addTocToManifest.tsx
```diff
@@ -0,0 +1,64 @@
...
+): Promise<ReadiumLink[]> => {
+  // iterate through each chapter
+  for (let i = 0; i < outline.length; i++) {
+    const chapter = outline[i];
+    const dest = chapter.dest;
+    if (dest && dest.length > 0) {
+      // get the page referance
+      const ref = dest[0];
+      const pageIndex = await pdf.getPageIndex(ref);
+      // TODO: Uncomment subchapter handling when goToPage is updated
```
kristojorg: In Open eBooks we have been trying to avoid TODO comments that don't link to a Jira ticket. I think it would be useful to make a Jira ticket in DRB for these TODOs and link them here

src/PdfReader/addTocToManifest.tsx
```diff
@@ -0,0 +1,64 @@
...
+  proxiedUrl: string
+): Promise<WebpubManifest> {
+  const pdfUrl = manifest.readingOrder[0].href;
+  try {
+    const pdf = await pdfjs.getDocument(proxiedUrl).promise;
+    const outline = await pdf.getOutline(); // get the TOC outline
+    let toc: ReadiumLink[] = [];
+    if (outline) {
+      toc = await addTocItem(pdf, outline, pdfUrl, []);
+    }
```
kristojorg: If it were me, I would use a `map` function here instead of a loop. I generally try to avoid passing variables around and mutating them (like `toc` in this case) as it can quickly become difficult to trace through the code and understand what is happening to a variable. 

This is how I would write this:
```ts

const toc: ReadiumLink[] = outline.map(chapter => {
  const dest = chapter.dest;
  if (dest && dest.length > 0) {
    // get the page referance
    const ref = dest[0];
    const pageIndex = await pdf.getPageIndex(ref);
    return {
        title: chapter.title,
        href: `${pdfUrl}#page=${pageIndex + 1}`,
        children: [], // TODO: replace with subChapters when goToPage is updated
      }
    } else {
      return undefined;
    }
}).filter((chapter): chapter is ReadiumLink => !!chapter)
```

src/PdfReader/index.tsx
```diff
@@ -195,12 +196,27 @@ const loadResource = async (resourceUrl: string, proxyUrl?: string) => {
...
 };
 
+/**
+ * Helper method to get hash page
+ * @param resourceUrl
+ * @returns
+ */
+const getHashPage = (resourceUrl: string): number => {
+  const hash = new URL(resourceUrl).hash;
+  return hash ? parseInt(hash.split('=')[1]) : 1;
```
kristojorg: Should we make sure that the has is `page=<number>` and not eg `#somethingelse=5`?

src/PdfReader/index.tsx
```diff
@@ -373,6 +389,21 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
+  React.useEffect(() => {
+    if (!manifest || !manifest.readingOrder || !manifest.readingOrder.length)
+      return;
+
+    if (isSinglePDF) {
+      const resourceUrl = getResourceUrl(
+        state.resourceIndex,
+        manifest.readingOrder
+      );
+      const proxiedUrl = `${proxyUrl}${encodeURIComponent(resourceUrl)}`;
```
kristojorg: Can we use the `loadResource` function here, which has built-in support for setups with or without a `proxyUrl`?

src/PdfReader/index.tsx
```diff
@@ -472,36 +503,71 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
     });
   }, [state.scale]);
 
+  /**
+   * TODO: Update to work with sub-chapter links
+   */
```
kristojorg: In Open eBooks we have been trying not to add TODO comments unless they link to specific Jira tickets. I think it would be good to continue that practice here so these TODOs don't get lost in the weeds

src/PdfReader/index.tsx
```diff
@@ -472,36 +503,71 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
-        type: 'SET_CURRENT_RESOURCE',
-        index: resourceIndex,
-        shouldNavigateToEnd: false,
-      });
-
-      if (manifest?.readingOrder && manifest?.readingOrder[resourceIndex]) {
-        const startPage = getStartPage(
-          manifest?.readingOrder[resourceIndex].href
-        );
+      } else {
```
kristojorg: My understanding of this change is that if the resource is a single resource pdf, we do not call `SET_CURRENT_RESOURCE` and we call `scrollIntoView` to scroll the user to the given page number. Given that understanding, I have two comments:

1. This code is almost the same as the code for a multi-resource pdf. It would be preferrable to have one general `goToHref` function that handles both cases simultaneously. This is how I imagine that functioning:
  - Get the index for the requested resource.
  - If the requested index is different from the current index, dispatch a `SET_CURRENT_RESOURCE` action (we can actually dispatch this no matter what, if the number doesn't change then nothing will happen).
  - Check if the requested resource has a `#page=<number>` hash parameter. If it does, dispatch a `NAVIGATE_PAGE` action to navigate us to that page.
  - If not, check instead fo the `startPage` hash. Navigate similarly to that page if it's present. 
2. These callbacks (`goToPage` in this case) should not have any side effects (like `scrollIntoView`). They should take input, dispatch actions, and optionally return something but not directly change state or manipulate the document in any way. The flow should be `goToPage("somewhere#page=5")` gets called. This in turn dispatches `NAVIGATE_PAGE` with `pageNum: 5`, which updates the local state. Then, a `useEffect` hook can detect this change and scroll the user to the right page. I admit this is a seemingly roundabout way to do something that is already being done (and working) here more directly. However, this separation of actions, state and effects becomes very important when things get more complex (which they inevitably do). If we navigate the user directly here, then in any function we write to navigate the user we have to remember to scroll them to the right location. Imagine we want to add a feature to take to user to their last known reading location. We make a function to get the last known location from storage and then navigate the user. We have to remember to call `scrolIntoView` in a similar manner there. 
Instead, my idea with this architecture is to do this declaratively. When we want to go somewhere, we just dispatch a `NAVIGATE_PAGE` action and can trust that the reducer will update the state and the effect will be called apropriately. 

I see that `NAVIGATE_PAGE` is already implemented, but I don't see any effect actually scrolling the user to the given page. I'm not sure how it is meant to work exactly but I will take a closer look and get back to you. We may need to add an effect to detect changes in the `pageNumber` state and call `scrollIntoView`.

src/PdfReader/index.tsx
```diff
@@ -655,3 +721,6 @@ function handleInvalidTransition(state: PdfState, action: PdfReaderAction) {
   );
   return state;
 }
+function href(href: any) {
```
kristojorg: Is this meant to be here?

src/PdfReader/index.tsx
kristojorg: Do we need to check if the key is `page` vs for example `#startPage=5`?

src/PdfReader/addTocToManifest.tsx
kristojorg: Do we need to use `any` here? I would rather explicitly type these if possible

kristojorg (changes requested):
This looks close! There are still a couple of changes I'd like to see before merging this though. I will make a PR to your branch to describe what I am thinking, and you can choose to accept those changes or not

kristojorg (approved):
Looks good to me! Thanks for your patience here and working with me : ) Feel free to merge and publish a new version at your convenience

PR #151, opened 2023-02-07
Kristo/pdf-toc

kristojorg:
This PR:

1. Extracts the reducer, types, and utility functions into separate files for easier understanding
2. Makes a `GO_FORWARD` `GO_BACKWARD` and `GO_TO_HREF` action.
3. Moves all action logic into the reducer (out of the callback) so it can be shared easier and centralized.
4. Makes an effect that scrolls the user to the correct page when the page number chanages. 
5. Updates `addTocToManifest` to avoid passing a mutable `toc` value to a separate function.

There is one more thing I would like to try, which is adding the toc to the manifest outside of the web reader. I will have to give this a go tomorrow

### DIFF
--- a/example/index.tsx
+++ b/example/index.tsx
@@ -9,7 +9,7 @@ import {
   Link,
   useParams,
 } from 'react-router-dom';
-import WebReader from '../src';
+import WebReader, { addTocToManifest } from '../src';
 import {
   ChakraProvider,
   Heading,
@@ -28,10 +28,12 @@ import readiumDefault from 'url:../src/HtmlReader/ReadiumCss/ReadiumCSS-default.
 import readiumAfter from 'url:../src/HtmlReader/ReadiumCss/ReadiumCSS-after.css';
 import Tests from './Tests';
 import { Injectable } from '../src/Readium/Injectable';
+import useSWR, { Fetcher } from 'swr';
 
 const origin = window.location.origin;
 
 const pdfProxyUrl = process.env.CORS_PROXY_URL as string | undefined;
+const pdfWorkerSrc = `${origin}/pdf-worker/pdf.worker.min.js`;
 
 const cssInjectables: Injectable[] = [
   {
@@ -99,11 +101,7 @@ const PdfReaders = () => {
   return (
     <>
       <Route path={`/pdf/single-resource-short`}>
-        <WebReader
-          webpubManifestUrl="/samples/pdf/single-resource-short.json"
-          proxyUrl={pdfProxyUrl}
-          pdfWorkerSrc={`${origin}/pdf-worker/pdf.worker.min.js`}
-        />
+        <SingleResourcePdf />
       </Route>
       <Route path={`/pdf/large`}>
         <WebReader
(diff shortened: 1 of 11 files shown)


example/index.tsx
```diff
@@ -159,6 +157,62 @@ const PdfReaders = () => {
...
+ */
+const fetchAndModifyManifest: Fetcher<string, string> = async (url) => {
+  const response = await fetch(url);
+  const manifest = await response.json();
+  const modifiedManifest = await addTocToManifest(
+    manifest,
+    getProxiedResource(pdfProxyUrl),
+    pdfWorkerSrc
+  );
+  const syntheticUrl = URL.createObjectURL(
```
jackiequach, line 193: Didn't know this was possible. Very cool!

src/PdfReader/addTocToManifest.tsx
```diff
@@ -1,58 +1,50 @@
...
+            title: chapter.title,
+            href: `${pdfUrl}#page=${pageIndex + 1}`,
+            children: [],
+          };
+          return link;
+        }
+      }
+      return undefined;
+    });
+    // await all the promises and filter any undefined values (tere was no chapter.dest or pageIndex)
```
jackiequach: typo in parenthesis

src/PdfReader/reducer.ts
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Utility function to generate state navigating us to a given resource and page.
+     * Used by multiple cases below.
+     */
+    function goToLocation(index: number, page = 1): PdfState {
+      // only set the resource to null if you're actually changing resources (not just
+      // navigating to a different page in the same resource)
+      const shouldResetResource = state.resourceIndex !== index;
+
+      // check if there is a #startPage in the href of the resource we are navigating to
```
jackiequach: change "#startPage" to "?startPage"

src/PdfReader/reducer.ts
```diff
@@ -0,0 +1,196 @@
...
+
+      case 'GO_BACKWARD': {
+        /**
+         * Navigate backward one page or to the end of the previous resource
+         * if at the beginning of the current resource. Do nothing at the
+         * beginning of the first resource.
+         */
+        // do nothing if we have not parsed the number of pages yet.
+        if (!state.numPages) return state;
+        const atStartOfResource = state.pageNumber === 1;
```
jackiequach: This will not work for multi-pdf resources since the start of the resource will be the value from ?startPage=X.

jackiequach (changes requested):
Overall changes look great! Left a few minor comments.

Also, when switching tabs for [Single-PDF Webpub](https://web-reader-git-kristo-pdf-toc-nypl.vercel.app/pdf/single-resource-short), the PDF seems to reload and the page resets to 1. I'm unsure where this issue may have been introduced.

src/PdfReader/reducer.ts
kristojorg: Very good point, I'll update this

example/index.tsx
```diff
@@ -159,6 +157,62 @@ const PdfReaders = () => {
...
+};
+
+/**
+ * - Fetches manifest
+ * - Adds the TOC to the manifest
+ * - Generates a syncthetic url for the manifest to be passed to
+ * web reader.
+ * - Returns the synthetic url
+ */
+const fetchAndModifyManifest: Fetcher<string, string> = async (url) => {
```
jackiequach, line 185: Should this method be moved and exported for consuming apps to use, or documented as a code snippet to be added in consuming apps?

kristojorg: That's not a bad idea... I was thinking we could leave the fetching to the consuming app in case they have particular ways they need to do it. I will update the README however with this code snippet : )

kristojorg:
Updated the README, LMK what you think @jackiequach : )

kristojorg:
@jackiequach I am unable to get cypress to run locally to debug the failing test... There seems to be some error with my M1 mac. Maybe cypress needs to be upgraded? Are you able to get it to run?

jackiequach:
@kristojorg I am able to run the cypress tests locally. multi-pdf/navigations.ts > Scrolling mode & TOC: should update page content after clicking on TOC link is failing. This is most likely due to skipping the interstitial pages on scroll mode and changing the content visible on the page.

samanthaandrews:
@jackiequach Since things are ramping up for Kristo on PJR, I'm thinking it makes sense for you to fix these Cypress tests and get this PR ready to merge. Let me know if you think this is something you can do. You can create a ticket for it or take over a the ticket that Kristo created if it exists.

We do need to figure out the issue with the Cypress tests not running for Kristo, but I think that task will have to wait.

kristojorg:
Thanks so much @jackiequach ! Feel free to merge this whenever you're ready

jackiequach (approved):

PR #153, opened 2023-04-07
Fix server rendering issues

kristojorg:
Ok I believe this should solve the errors we were seeing in DRB previously. This PR

1. Changes the fullscreen code to not depend on `document` being defined when the code is imported on the server.
2. Swaps out our custom `esbuild` build script for `tsup`, which uses `esbuild` under the hood. This should have saner defaults, work better and incur less maintenance (I was getting build errors previously)
3. I had to change how the `AnnotationLayer.css` file was imported. It now gets bundled into an `index.css` file in `dist`. I'm not sure how DRB previously included that file, but if you let me know, I can update this to be at the same place.


samanthaandrews (approved):
@jackiequach take a look at this and let me know what you think!

jackiequach (approved):
Looks good to me!

kristojorg:
@jackiequach is this still an issue in DRB? If so, do you want to merge this and publish a release and update DRB to use it?

jackiequach:
@kristojorg Yes, this is needed for DRB.

kristojorg:
Ok I will merge this and publish a release then

kristojorg:
This has been released as 4.3.1

PR #154, opened 2023-10-10
Chore/remove-old-axis

kristojorg:
This removes the axisnow package which is now deprecated. It was only used in the example app as an example, but we can't do that anymore as it is now internal to ereading-clients


Toxiapo (approved):
LGTM! There's also the encrypted assets in the samples folder, but I am ok with keeping them for reference.

samanthaandrews (approved):

PR #155, opened 2024-02-22
OE-742: Add Skip to Content link

jackiequach:
Adds "Skip to Main Content" link to go directly to the book content. The new component is based on the [DS SkipNavigation](https://nypl.github.io/nypl-design-system/reservoir/v2/?path=/docs/components-navigation-skipnavigation--docs) component. Replaces iframe ids with "mainContent".

### DIFF
--- a/cypress/integration/multi-pdf/navigations.ts
+++ b/cypress/integration/multi-pdf/navigations.ts
@@ -35,7 +35,7 @@ describe('Multi PDF navigation', () => {
   it('should navigate forward and backwards with page buttons', () => {
     cy.findByText('Anthropology without Informants').should('be.visible');
     cy.findByRole('button', { name: 'Next Page' }).click();
-    cy.get('#iframe-wrapper')
+    cy.get('#mainContent')
       .find('div[class="react-pdf__Page__textContent"]')
       .children()
       .should('have.length', 0);
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -170,7 +170,7 @@ Cypress.Commands.add(
       },
     });
     cy.wait('@pdf', { timeout: 50000 });
-    cy.get('#iframe-wrapper')
+    cy.get('#mainContent')
       .find('div[class="react-pdf__Page__textContent"]', { timeout: 10000 })
       .should('have.attr', 'style');
   }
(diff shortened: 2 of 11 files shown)


jackiequach:
Cypress tests are failing due to DRB QA being behind VPN. The PDF resources go through DRB.

samanthaandrews:
I'm going to message Ayan and Anthony about the DRB QA issue. Seems like this should have a resolution by now since it's been several weeks that DRB QA has been behind a VPN.

I'm going to let @kristojorg and @Toxiapo and @oliviawongnyc review this PR since they are more familiar with the code base.

kristojorg (approved):
Changes look good to me! Thanks for adding this : )

samanthaandrews:
I'm ok with merging this with the tests failing for now.

@kristojorg do you know why we made some of the Cypress tests run against DRB QA site instead of locally? I'm assuming it has something to do with the PDF proxy.

@jackiequach said that the VPN on DRB QA can be taken down once the limited access feature is deployed to prod, which should be happening in a few weeks, so I'm going to wait to make tickets to fix these tests to see if the issue resolves on its own.

Toxiapo:
> I'm ok with merging this with the tests failing for now.
> 
> @kristojorg do you know why we made some of the Cypress tests run against DRB QA site instead of locally? I'm assuming it has something to do with the PDF proxy.
> 
> @jackiequach said that the VPN on DRB QA can be taken down once the limited access feature is deployed to prod, which should be happening in a few weeks, so I'm going to wait to make tickets to fix these tests to see if the issue resolves on its own.

I don't remember why either, I think the proxy was already available for us so we just used it for our tests, I wonder if we should just host the proxy service our own in vercel or something if that would make things easier.

samanthaandrews:
> I don't remember why either, I think the proxy was already available for us so we just used it for our tests, I wonder if we should just host the proxy service our own in vercel or something if that would make things easier.

Thanks @Toxiapo! I think hosting the proxy service in Vercel is a good idea. If the DRB site doesn't move off of the VPN within the next couple of weeks I'll make tickets to implement this.

jackiequach:
I don't have the perms to bypass the merge requirements.

PR #157, opened 2024-04-10
OE-818: Update React, Node, Chakra, and other packages

jackiequach:
[OE-818](https://jira.nypl.org/browse/OE-818)

Tests for single- and multi- pdf are failing due to DRB being behind the VPN

- Updates to React 18, Node 20, Chakra 2.8
- Updates react-pdf to 7.7.1 and the associated pdf-worker files (copied from the pdfjs-dist node_modules as per react-pdf setup instructions)
- Updates broken Cypress tests
- Removes @testing-library/react-hooks since it is built into React Testing Library


package.json
```diff
@@ -34,14 +34,6 @@
     "release": "release-it --verbose",
     "release:alpha": "release-it --verbose --preRelease=alpha"
   },
-  "peerDependencies": {
-    "@chakra-ui/react": "^1.8.7 || 2.x",
-    "@emotion/react": "^11.8.2",
-    "@emotion/styled": "^11.8.1",
-    "react": "^16.8.0 || 17.x || 18.x",
-    "react-dom": "^16.8.0 || 17.x || 18.x",
-    "swr": "^2.0.3"
```
kristojorg: I think these should stay as peer dependencies rather than production dependencies because it allows the dependency to be satisfied by an installing repo having a similar version

kristojorg (approved):
Looks good! Just the one comment about the peer dependencies

package.json
samanthaandrews: I do think we can keep these, but they should be updated to require react 18 and chakra v2. Using chakra v2 in the consuming app while the web reader was on chakra v1 was causing some very strange styling issues.

kristojorg: Oh yes true. The updates were good, just should stay peer instead of production dependencies

kristojorg (approved):

PR #158, opened 2024-04-16
NO-REF: Update version to 4.3.3 in package.json

jackiequach:
Update version to 4.3.3 in package.json

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@nypl/web-reader",
-  "version": "4.3.2",
+  "version": "4.3.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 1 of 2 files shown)


samanthaandrews:
iirc i think the process for updating the package version number is done through the instructions here https://github.com/NYPL-Simplified/web-reader?tab=readme-ov-file#publishing-to-npm

is this correct, @kristojorg ? I don't know if I've ever even done a web-reader release before 😬

jackiequach:
It is done there, but since I don't have permissions for the main branch it skipped that step.

samanthaandrews:
> It is done there, but since I don't have permissions for the main branch it skipped that step.

I made you an admin... maybe that will allow you?!

samanthaandrews (approved):
I don't think merging this will make the release show up in GitHub, which is annoying. But I do see that 4.3.3 has been released on NPM!! 🎉 https://www.npmjs.com/package/@nypl/web-reader

PR #160, opened 2024-04-19
NO-REF: Add export for type and fix warnings

jackiequach:
- Add export for ReadiumLink and WebpubManifest types to be used in consuming apps
- Fix warnings from browser for deprecated functions

### DIFF
--- a/src/ui/nypl-base-theme/foundations/breakpoints.ts
+++ b/src/ui/nypl-base-theme/foundations/breakpoints.ts
@@ -1,10 +1,10 @@
 // theme.js
-import { createBreakpoints } from '@chakra-ui/theme-tools';
-
-export default createBreakpoints({
+const breakpoints = {
   sm: '20em', // 320px
   md: '38em', // 600px
   lg: '60em', // 960px
   xl: '80em', // 1280px
   '2xl': '96em',
-});
+};
+
+export default breakpoints;
--- a/tests/jest.setup.ts
+++ b/tests/jest.setup.ts
@@ -1,4 +1,2 @@
 import '@testing-library/jest-dom';
 import 'jest-axe/extend-expect';
-
-jest.setTimeout(15000);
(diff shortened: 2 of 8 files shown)


Toxiapo (approved):

src/ui/nypl-base-theme/foundations/breakpoints.ts
```diff
@@ -1,10 +1,10 @@
 // theme.js
-import { createBreakpoints } from '@chakra-ui/theme-tools';
-
-export default createBreakpoints({
```
kristojorg, line 4: Just wondering, why remove the `createBreakpoints` function? It's only really there to aid in type checking

tests/jest.setup.ts
```diff
@@ -1,2 +1,4 @@
 import '@testing-library/jest-dom';
 import 'jest-axe/extend-expect';
+
+jest.setTimeout(15000);
```
kristojorg: Does this need to be here?

kristojorg (commented):
Just two questions, but otherwise looks great!

src/ui/nypl-base-theme/foundations/breakpoints.ts
jackiequach: There was a warning that this is being deprecated by Chakra soon and to just pass the breakpoints object to the theme.

tests/jest.setup.ts
jackiequach: I moved the timeout to the specific tests (accessibility checks).

src/ui/nypl-base-theme/foundations/breakpoints.ts
kristojorg: Sweet thanks for the explanation!

kristojorg (approved):
Looks good!

PR #162, opened 2024-06-14
SFR-1969: Implement getContent in the pdf reader

jackiequach:
Related to the work in [Jira Ticket](http://jira.nypl.org/browse/SFR-1969)

Implements getContent custom fetcher for the pdf reader

### DIFF
--- a/src/PdfReader/index.tsx
+++ b/src/PdfReader/index.tsx
@@ -13,7 +13,12 @@ import {
   DEFAULT_SHOULD_GROW_WHEN_SCROLLING,
 } from '../constants';
 import LoadingSkeleton from '../ui/LoadingSkeleton';
-import { fetchAsTxt, getResourceUrl, loadResource, SCALE_STEP } from './lib';
+import {
+  getContentDefault,
+  getResourceUrl,
+  loadResource,
+  SCALE_STEP,
+} from './lib';
 import { makePdfReducer } from './reducer';
 
 /**
@@ -35,7 +40,7 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
     webpubManifestUrl,
     manifest,
     proxyUrl,
-    getContent = fetchAsTxt,
+    getContent = getContentDefault,
     injectablesReflowable,
     injectablesFixed,
     height = DEFAULT_HEIGHT,
--- a/src/PdfReader/index.tsx
+++ b/src/PdfReader/index.tsx
@@ -1,6 +1,6 @@
 import { Document, PageProps, pdfjs } from 'react-pdf';
 import * as React from 'react';
-import { ReaderArguments, ReaderReturn } from '../types';
+import { ReaderReturn } from '../types';
 import { Flex } from '@chakra-ui/react';
 import useMeasure from './useMeasure';
 import ChakraPage from './ChakraPage';
@@ -13,13 +13,9 @@ import {
   DEFAULT_SHOULD_GROW_WHEN_SCROLLING,
 } from '../constants';
 import LoadingSkeleton from '../ui/LoadingSkeleton';
-import {
-  getContentDefault,
-  getResourceUrl,
-  loadResource,
-  SCALE_STEP,
-} from './lib';
+import { getContentDefault, getResourceUrl, SCALE_STEP } from './lib';
 import { makePdfReducer } from './reducer';
+import { PdfReaderArguments } from './types';
 
 /**
  * The PDF reader
@@ -30,7 +26,7 @@ import { makePdfReducer } from './reducer';
  * @param args T
  * @returns
  */
-export default function usePdfReader(args: ReaderArguments): ReaderReturn {
+export default function usePdfReader(args: PdfReaderArguments): ReaderReturn {
   // use a passed in src for the pdf worker
   if (args?.pdfWorkerSrc) {
     pdfjs.GlobalWorkerOptions.workerSrc = args.pdfWorkerSrc;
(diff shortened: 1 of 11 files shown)


src/PdfReader/index.tsx
```diff
@@ -104,17 +104,25 @@ export default function usePdfReader(args: ReaderArguments): ReaderReturn {
...
+
+      loadResource(resourceUrl, proxyUrl).then((data) => {
+        dispatch({
+          type: 'RESOURCE_FETCH_SUCCESS',
+          resource: { data },
+        });
       });
-    });
-  }, [state.resourceIndex, manifest, proxyUrl]);
+    };
```
kristojorg: I'm a little confused by what `getContent` does here and it's role in relation to `loadResource`. It looks like the default `getContent` (`fetchAsTxt`) in this case doesn't fetch anything, but just returns the url passed in. Can you give me an example of how you might use this for authenticated content?

In the HTML Reader, we would use it like this:
```ts
function fetchWithAuth(resourceUrl: string) {
  const respose = fetch(resourceUrl, {
    headers: {
      Authentication: "bearer MyToken"
    }
  })
  const text = await response.text();
  return text;
}
function ReaderWithAuth() {
  /*
  ... some other bits
  */
 return <HtmlReader getContent={fetchWithAuth} />
}
```
In the above, the `getContent` we pass in will fetch the content however is necessary for the app (adding headers, or processing the result in different ways before returning it).

But I'm not sure how this would work in the PDFReader, because in that case it looks like `getContent` just returns a url, which is then fetched and processed by `loadResource`.

My suggestion would be that the function performed by `loadResource` right now can be instead performed by `getContent`. That is, `getContent` in the PDF reader should take a `url` and return a `Promise<Uint8Array>`. Then, the consuming app can implement it however they see fit for their use case (adding headers or processing content however they need).

Does that make sense? Maybe I'm missing something. It might be helpful if you could write out an example of how DRB could use this new `getContent` prop to fetch the content it needs

jackiequach: Here is an example of how `getContent` will be used:
```
const getContent = async (linkUrl: string) => {
  const res = await fetch(linkUrl.toString(), {
    method: "GET",
    headers: {
      Authorization: `Bearer access_token`,
    },
  });
  return res.url;
};
// ...
<WebReader
  webpubManifestUrl={manifestUrl}
  proxyUrl={proxyUrl}
  pdfWorkerSrc={pdfWorkerSrc}
  headerLeft={<BackButton />}
  injectablesFixed={injectables}
  getContent={isLimitedAccess ? getContent : null}
/>
```
The link that will be passed to `getContent` would be a `/fulfill` endpoint that redirects to the returned resource URL.  This resource URL is then processed by the PDFReader's `loadResource`.

The default `getContent` (`fetchAsTxt`)  is meant to keep the functionality of `loadResource` as is. I could rename it to reflect its behavior.

kristojorg: I see, so the contents of the books are actually publicly accessible on the web, but you just have to know the url to call, is that right? Couldn't a user see that url in their browser and share it then?

In any case, the API for `getContent` is that it is expected to return the resource itself, which you should be able to implement here simply enough by just fetching the url instead of returning it, and returning the Uint8Array from `getContent`. Does that make sense?

jackiequach: The contents require authentication, but the url can be shared by the user post-auth. The backend team is aware of this and is looking at different solutions. 

Yes, I'll update the `getContent` definition for this.

src/types.ts
```diff
@@ -87,16 +87,19 @@ export type ActiveReader = PDFActiveReader | HTMLActiveReader;
 export type ReaderReturn = InactiveReader | LoadingReader | ActiveReader;
 
 // should fetch and decrypt a resource
-export type GetContent = (href: string) => Promise<string>;
+export type GetContent<T extends string | Uint8Array> = (
```
jackiequach, line 90: Let me know if there is a better way to update the type definitions that I may have overlooked.

src/PdfReader/lib.ts
```diff
@@ -33,6 +33,16 @@ export const loadResource = async (
...
+/**
+ * Fetches a resource url as text
+ */
+export async function getContentDefault(
+  resourceUrl: string,
+  proxyUrl?: string
+): Promise<Uint8Array> {
+  return loadResource(resourceUrl, proxyUrl);
+}
+
```
kristojorg: My last thought is - if this just calls loadResource, should we just use that function directly and rename it `getContentDefault` or `fetchAsUint8Array`?

src/types.ts
kristojorg: I think this is okay for now. I think the only other option would be splitting out separate types/arguments/returns for the `HtmlReader` and `PdfReader`, which is probably more work than it's worth right now

src/useWebReader.tsx
```diff
@@ -67,6 +68,8 @@ export default function useWebReader(
   const readerType = getReaderType(
     manifest ? manifest.metadata.conformsTo : null
   );
+  const getContentHtml = getContent as GetContent<string> | undefined;
+  const getContentPdf = getContent as GetContent<Uint8Array> | undefined;
```
kristojorg, line 72: This isn't ideal, but I think it's okay for right now. In the future I'm thinking we might just want to not have a `useWebReader` and instead only export a `useHtmlReader` and `usePdfReader` separately. Seems like it might be simpler for these kinds of types, and I don't forsee that many cases when you would need to use both simultaneously

kristojorg (approved):
Looks good! Just the minor comment about `loadResource` and `getContent` doing the same thing. Thanks for the changes!

PR #163, opened 2024-06-27
SFR-1944: Add error state for PDFReader

jackiequach:
Adds an error state for the PDFReader. The error state is set if the document fails to load and throws the error when updated.


kristojorg (approved):
Looks good to me!

PR #164, opened 2025-03-13
OE-1455: Set up Playwright

clarissarichard:
This PR adds Playwright to the web-reader repo and creates a test to open the direct link to an epub and verify that the Back button is visible.

### DIFF
--- a/playwright/tests/test.spec.ts
+++ b/playwright/tests/test.spec.ts
@@ -1,9 +1,6 @@
 import { test, expect } from '@playwright/test';
 
 test('Open moby-epub3', async ({ page }) => {
-  await page.goto('http://localhost:1234/html/moby-epub3');
-  //await expect(page.getByRole('link', { name: 'Back' })).toBeVisible(); // does not work
-  //await expect(page.getByRole('paragraph', { name: 'Back' })).toBeVisible(); // does not work
-  //await expect(page.getByText('Back')).toBeVisible(); // works but could be found in epub
-  await expect(page.getByLabel('Return to Homepage')).toBeVisible(); // works
+  await page.goto('/html/moby-epub3');
+  await expect(page.getByLabel('Return to Homepage')).toBeVisible();
 });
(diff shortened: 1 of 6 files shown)


playwright/tests/test.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { test, expect } from '@playwright/test';
+
+test('Open moby-epub3', async ({ page }) => {
+  await page.goto('http://localhost:1234/html/moby-epub3');
+  //await expect(page.getByRole('link', { name: 'Back' })).toBeVisible(); // does not work
+  //await expect(page.getByRole('paragraph', { name: 'Back' })).toBeVisible(); // does not work
+  //await expect(page.getByText('Back')).toBeVisible(); // works but could be found in epub
+  await expect(page.getByLabel('Return to Homepage')).toBeVisible(); // works
```
samanthaandrews: nice `getByLabel` is good because it also tests that the site is accessible

playwright/tests/test.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { test, expect } from '@playwright/test';
+
+test('Open moby-epub3', async ({ page }) => {
+  await page.goto('http://localhost:1234/html/moby-epub3');
+  //await expect(page.getByRole('link', { name: 'Back' })).toBeVisible(); // does not work
+  //await expect(page.getByRole('paragraph', { name: 'Back' })).toBeVisible(); // does not work
+  //await expect(page.getByText('Back')).toBeVisible(); // works but could be found in epub
```
samanthaandrews: let's just delete these 3 lines

playwright/tests/test.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { test, expect } from '@playwright/test';
+
+test('Open moby-epub3', async ({ page }) => {
+  await page.goto('http://localhost:1234/html/moby-epub3');
```
samanthaandrews: i think we can put `http://localhost:1234` into the Playwright config file somewhere https://playwright.dev/docs/test-configuration

samanthaandrews (changes requested):
just a few small changes. looks great!

samanthaandrews (approved):
🚀

PR #165, opened 2025-03-14
OE-1646: Write reader settings tests

clarissarichard:
This PR includes Playwright tests to confirm the reader settings are functional.

It currently tests the following in an HTML pub and a PDF pub:
- All settings buttons are visible
- Displays default settings
- Changes to font, background color, text size, and pagination style
- Changes settings in full screen
- Zooms in and out
- Maintains changed settings
- Resets settings


playwright/pageobjects/web-reader.page.ts
```diff
@@ -0,0 +1,139 @@
...
+        .scrollIntoViewIfNeeded();
+    } else if (currentURL.includes('/html/moby-epub3')) {
+      await this.tocButton.click();
+      await this.page
+        .getByText('EXTRACTS (Supplied by a Sub-Sub-Librarian).')
+        .click();
+      await this.page
+        .frameLocator('#mainContent')
+        .getByText('WHALE SONG')
+        .scrollIntoViewIfNeeded();
```
clarissarichard: It doesn't appear to scroll to the last line of this chapter, so not sure that I'm getting into the iframe correctly. There's also two places where "WHALE SONG" could be found: `iframe`'s `srcdoc` attribute and in `iframe > html > body > div` - is one preferable over the other?

playwright/pageobjects/web-reader.page.ts
```diff
@@ -0,0 +1,139 @@
...
+      .locator('html')
+      .getAttribute('style');
+    console.log(htmlElement);
+    // htmlElement.evaluate((el) => {
+    //   return window.getComputedStyle(el).getPropertyValue('--USER__fontSize');
+    // });
+  }
+
+  // fix
+  async scrollDown(): Promise<void> {
```
clarissarichard, line 133: Is `Promise<void>` the correct return type here?

playwright/pageobjects/web-reader.page.ts
```diff
@@ -0,0 +1,139 @@
...
+    //     expectedValueX: string,
+    //     expectedValueY: string
+    //   ): void => {
+    //     this.page.window().then((win) => {
+    //       const styles = win.getComputedStyle($elm[0]);
+    //       const scaleX = styles.getPropertyValue('--chakra-scale-x');
+    //       const scaleY = styles.getPropertyValue('--chakra-scale-y');
+    //       expect(scaleX).to.eq(expectedValueX);
+    //       expect(scaleY).to.eq(expectedValueY);
+    //     });
```
clarissarichard: I pulled all this from the Cypress tests and commented it out because it felt a bit daunting to translate to Playwright at the moment. I see a `--scale-factor` value in the `style` attribute in the `react-pdf__Page css-10sejjj` class, and I'm envisioning comparing the values before and after zooming in/out. Just need to figure out the iframe first.

playwright/pageobjects/web-reader.page.ts
Toxiapo: yup, that's correct! You can set the return type to `void` for any functions that are not returning anything, and if it's `async` function, it's `Promise<void>`

Toxiapo: We can also refactor this method a bit. I don't think checking the URL is a good idea here, maybe   break this down into two functions, `scrollDownPdf`,  `scrollDownHtml`, what do you think?

Toxiapo: Another idea is to have two classes,  one for `PdfReaderPage` and one for `HtmlReaderPage` and separate the logics to the corresponding media type, which might be a little cleaner.

playwright/pageobjects/web-reader.page.ts
Toxiapo: hmm yea that's strange, the latter feels a bit strange to me. I will have to double check on this!

playwright/pageobjects/web-reader.page.ts
```diff
@@ -0,0 +1,139 @@
...
+    this.paginatedStyle = page.getByText('Paginated', { exact: true });
+    this.scrollingStyle = page.getByText('Scrolling', { exact: true });
+
+    // footer
+    this.previousPageButton = page.getByLabel('Previous Page');
+    this.nextPageButton = page.getByLabel('Next Page');
+  }
+
+  // hopefully better handles slow load time (use load or networkidle)
+  async loadPage(gotoPage: string) {
```
Toxiapo: return type ` Promise<WebReaderPage> `

playwright/pageobjects/web-reader.page.ts
```diff
@@ -0,0 +1,139 @@
...
+    // header
+    this.backButton = page.getByLabel('Return to Homepage');
+    this.tocButton = page.getByLabel('Table of Contents');
+    this.firstChapter = page.getByRole('menuitem').first();
+    this.lastChapter = page.getByRole('menuitem').last();
+    this.settingsButton = page.getByRole('button', {
+      name: 'Settings',
+      exact: true,
+    });
+    this.fullScreenButton = page.getByText('Full screen', { exact: true });
```
Toxiapo: For button, it should be easier to use the role label. ` page.getByRole('button', { name: 'Toggle full screen' })`

playwright/pageobjects/web-reader.page.ts
Toxiapo: Yup! This is the only way to check the changes for now.What we wanted to test here is whether the setting alters the state of these attributes.

playwright/pageobjects/web-reader.page.ts
clarissarichard: Ooh, this one does work with `getByRole`! I'll update it. 

I originally had `getByRole` with the other settings buttons, but it wasn't able to click them for some reason. When I ran `npx playwright test --ui` and looked at the step, it attempted to click the very corner of the button. Perhaps it only affected radio buttons.

playwright/pageobjects/web-reader.page.ts
```diff
@@ -0,0 +1,172 @@
...
+    this.previousPageButton = page.getByLabel('Previous Page');
+    this.nextPageButton = page.getByLabel('Next Page');
+  }
+
+  // hopefully better handles slow load time (use load or networkidle)
+  async loadPage(gotoPage: string): Promise<void> {
+    await this.page.goto(gotoPage, { waitUntil: 'load' });
+  }
+
+  async getIframe() {
```
Toxiapo: return type `Promise<Locator>`

playwright/pageobjects/web-reader.page.ts
```diff
@@ -0,0 +1,172 @@
...
+  async getTextSize() {
+    const htmlElement = this.page
+      .frameLocator('#mainContent')
+      .locator('html')
+      .getAttribute('style');
+    console.log(htmlElement);
+    // htmlElement.evaluate((el) => {
+    //   return window.getComputedStyle(el).getPropertyValue('--USER__fontSize');
+    // });
+  }
```
Toxiapo: This is very close! I think this would work if we just do
```suggestion
  async getTextSize(): Promise<undefined | string> {
    return (await this.getIframe()).evaluate((el) => {
      return window.getComputedStyle(el).getPropertyValue('--USER__fontSize');
    });
  }
```

playwright/tests/reader-settings.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  test('Displays default settings', async ({ page }) => {
+    const webReaderPage = new WebReaderPage(page);
+    await webReaderPage.loadPage(
+      'https://nypl-web-reader.vercel.app/html/moby-epub3'
+    ); // temporary
+    await webReaderPage.settingsButton.click();
+    await expect(webReaderPage.defaultFont).toBeChecked();
+    await expect(webReaderPage.whiteBackground).toBeChecked();
+    //await expect(webReaderPage.getTextSize).toBe('100%'); // expect --USER__fontSize: 100%
```
Toxiapo: The `getTextSize` is an async function, so this would need to be awaited and invoked.

```suggestion
    await expect(await webReaderPage.getTextSize()).toBe('100%'); 
```

playwright/pageobjects/web-reader.page.ts
clarissarichard: However, it looks like `aria-label='Toggle full screen'` whether the page is in full screen or not. I would need a couple more locators to determine if the text says `Full screen` or `Full screen exit`.

Toxiapo (commented):
I haven't finished reviewing everything yet, but I'll pick this up tomorrow. This looks really good so far, and iframe seems better in PW compare to Cypress!

playwright/tests/reader-settings.spec.ts
clarissarichard: Omg this works now!! Thanks so much!

playwright/pageobjects/web-reader.page.ts
Toxiapo: Yea let's use the text selector for this! I think this might be a accessibility bug and we should change the label

Toxiapo:
@clarissarichard Looks like the iframe problem is fix?! But please feel free to ping me if there's anything else I can check to help unblock you.

playwright/pageobjects/web-reader.page.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  async zoomIn(): Promise<void> {
+    const beforeScaleFactor = await this.getZoomValue();
+    console.log('before: ', beforeScaleFactor);
+    this.settingsButton.click();
+    this.zoomInButton.click();
+    const afterScaleFactor = await this.getZoomValue();
+    console.log('after: ', afterScaleFactor); // currently the same value as beforeScaleFactor
+    expect(beforeScaleFactor).toBeGreaterThan(afterScaleFactor);
+  }
```
clarissarichard, line 169: @Toxiapo Forgot to add this comment! Is there a way to "reset" `getZoomValue()`? I'm getting the scale factor before and after zooming in and `beforeScaleFactor` and `afterScaleFactor` are returning the same value.

Toxiapo: hmm, there may be race condition here where you were grabbing the value and clicking the button at the same time.

Could you try to add the `await`  keyword to the two button clicks and let me know if that does anything?

```
await this.settingsButton.click();
await this.zoomInButton.click();
```

clarissarichard: Ah you're so right! It works now, thank you! When in doubt, use `await`.

playwright/pageobjects/web-reader.page.ts
```diff
@@ -0,0 +1,186 @@
...
+  async scrollDown(): Promise<void> {
+    await this.tocButton.click();
+    await this.page
+      .getByText('EXTRACTS (Supplied by a Sub-Sub-Librarian).')
+      .click();
+    await this.page
+      .locator('iframe[title="Moby-Dick"]')
+      .contentFrame()
+      .getByText('—WHALE SONG.')
+      .scrollIntoViewIfNeeded();
```
clarissarichard: @Toxiapo This is the other function I'm having trouble with. It doesn't appear to scroll down to the text at the bottom of the page per the Playwright UI tool, even though it's resolving to the correct element:
Scroll down in HTML pub
I've confirmed `scrollDown()` works in a PDF pub, but it's struggling on an HTML pub.

playwright/tests/html-reader-settings.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+    await htmlReaderPage.fullScreenButton.click();
+    await htmlReaderPage.exitFullScreenButton.click();
+    await expect(htmlReaderPage.fullScreenButton).toBeVisible();
+  });
+
+  test('Change settings in full screen', async ({ page }) => {
+    const htmlReaderPage = new HtmlReaderPage(page);
+    await htmlReaderPage.loadPage('/html/moby-epub3');
+    await htmlReaderPage.fullScreenButton.click();
+    await htmlReaderPage.changeSettings();
```
samanthaandrews, line 175: Should we add an `expect` statement somewhere in this test?

playwright/tests/pdf-reader-settings.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    await pdfReaderPage.fullScreenButton.click();
+    await pdfReaderPage.exitFullScreenButton.click();
+    await expect(pdfReaderPage.fullScreenButton).toBeVisible();
+  });
+
+  test('Change settings in full screen', async ({ page }) => {
+    const pdfReaderPage = new PdfReaderPage(page);
+    await pdfReaderPage.loadPage('/pdf/collection');
+    await pdfReaderPage.fullScreenButton.click();
+    await pdfReaderPage.changeSettings();
```
samanthaandrews, line 72: Same as above: Should we add an expect here?

samanthaandrews (approved):
Looks so good! Great progress. Feels good to even have just this set of tests converted. Just a couple comments.

playwright/tests/html-reader-settings.spec.ts
clarissarichard: I had `expect` statements in `changeSettings()`, but I will pull them out and move them into this test!

PR #168, opened 2025-04-07
SFR-2592: Update <Document> type to support File type from react-pdf

Toxiapo:
This PR updates the file prop type to align with react-pdf, allowing the consumer app to choose a different downloading strategy. By default, when the file prop is provided as a URL string, React-pdf will download and render the PDF content in chunks.

Additionally, I've added a new route, `Fetch PDF URL - Webpub`, which eliminates the need to construct a TOC for the manifest, this step will eventually be handled by the backend. I’ve also added a custom `getContent` function to return a URL string to replace the default function that is currently returns Uinit8Array.

## Tested on a 1.2G linearized PDF:
Before:
Before
After:
After

### DIFF
--- a/example/index.tsx
+++ b/example/index.tsx
@@ -241,7 +241,6 @@ const LoadPdfWithUrl = () => {
     async (url: string) => {
       const response = await fetch(url);
       const manifest = await response.json();
-      console.log(manifest);
       const syntheticUrl = URL.createObjectURL(
         new Blob([JSON.stringify(manifest)])
       );
--- a/example/index.tsx
+++ b/example/index.tsx
@@ -107,6 +107,9 @@ const PdfReaders = () => {
       <Route path={`/pdf/single-resource-short`}>
         <SingleResourcePdf />
       </Route>
+      <Route path={`/pdf/fetch-url`}>
+        <LoadPdfWithUrl />
+      </Route>
       <Route path={`/pdf/use-pdf-reader-hook`}>
         <UsePdfReader
           webpubManifestUrl="/samples/pdf/single-resource-short.json"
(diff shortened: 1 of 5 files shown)


src/PdfReader/lib.ts
```diff
@@ -19,7 +19,7 @@ export const getResourceUrl = (
 export const fetchAsUint8Array = async (
```
jackiequach, line 19: This should be renamed to match the new return type or simply to fetchPdf

example/index.tsx
```diff
@@ -228,6 +231,51 @@ const SingleResourcePdf = () => {
...
+ * This is an example for passing in a custom getContent callback function to the webreader hook
+ * where the pdf `file` type accepts an `url` string instead of the default BinaryData
+ */
+const LoadPdfWithUrl = () => {
+  const { data: modifiedManifestUrl, isLoading } = useSWR<string>(
+    '/samples/pdf/single-resource-short.json',
+    async (url: string) => {
+      const response = await fetch(url);
+      const manifest = await response.json();
+      console.log(manifest);
```
jackiequach: not sure if this is left here intentionally

jackiequach (approved):
looks good, just some minor non-blocking comments

example/index.tsx
Toxiapo: oops - deleting!

src/PdfReader/lib.ts
Toxiapo: I think the function name is correct here because we still want the default function to return Unit8Array!

kristojorg (approved):
Looks good to me!

Toxiapo:
I’ll close this for now—happy to chat about it anytime if it’s still relevant for your team! cc @jackiequach @alea12

PR #191, opened 2026-02-12
SCHOL-314: Re-enable and update pdf tests

jackiequach:
[SCHOL-314](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-314)

Base branch is the design implementation branch for UI changes.

- Re-enables PDF tests with a new webpub manifest using a PDF hosted in our S3 bucket.
- Adds a test for page input updates.
- Removes PDF settings (paginated/scrolling mode) tests since the button no longer exists.
- Updates addTocToManifest to handle PDFs with no outlines.

[SCHOL-314]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/PdfReader/addTocToManifest.tsx
```diff
@@ -21,6 +21,10 @@ export default async function addTocToManifest(
     pdfjs.GlobalWorkerOptions.workerSrc = pdfWorkerSrc;
     const pdf = await pdfjs.getDocument(pdfData).promise;
     const outline = await pdf.getOutline(); // get the TOC outline
+    if (!outline || !Array.isArray(outline) || outline.length === 0) {
+      return manifest;
+    }
+
```
alea12: Could you elaborate why this was necessary? `getOutline()` seems to be typed as an array already.

playwright/tests/pdf-reader-buttons.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    await pdfReaderPage.loadPub('/pdf/single-resource-short');
+    await expect(pdfReaderPage.fullScreenButton).toBeVisible();
+    await expect(pdfReaderPage.zoomInButton).toBeVisible();
+    await expect(pdfReaderPage.zoomOutButton).toBeVisible();
+  });
+
+  test('Zoom in', async ({ page }) => {
+    const pdfReaderPage = new PdfReaderPage(page);
+    await pdfReaderPage.loadPub('/pdf/single-resource-short');
+    await expect(await pdfReaderPage.zoomIn()).toBeTruthy;
```
alea12: should this be a function call without `await`? same goes for L22.

```ts
expect(await pdfReaderPage.zoomIn()).toBeTruthy();
```

src/PdfReader/addTocToManifest.tsx
jackiequach: You're right, this is not needed. I thought getOutline could return null/undefined.

playwright/tests/pdf-reader-buttons.spec.ts
jackiequach: removed the outer assertion call since the zoomIn function includes assertions

alea12 (approved):
Looking good!

PR #193, opened 2026-02-23
NOREF: Fix compatibility issues with Next.js 

jackiequach:
- Downgrades react-pdf 9.2.1 to avoid pdfjs (DOMMatrix) SSR/build failures in Next.js. This is a [known issue](https://github.com/wojtekmaj/react-pdf/issues/2039) which may be resolved in newer Next.js (untested).
- To prevent the runtime error, [CSS Imported by a Dependency]( https://nextjs.org/docs/messages/css-npm), the consuming app will need to import the rect-pdf CSS files.


alea12 (approved):
Thank you for updating the docs as well!

PR #197, opened 2026-03-03
NOREF: Fix PDF scrolling and programmatic navigation bugs

jackiequach:
- Fixes issue when navigating with the arrows will scroll the page to the top of the reader
    - Updated to scroll only within the reader component
- Fixes an issue where the default height of the reader is not set to the height of the browser
- Minor optimizations to `onInView` to ignore off-screen pages. Adds a programmatic navigation delay that prevents `goForward`, `goBackward`, and `goToPage` from triggering `onInView`, which sometimes caused the reader to jump pages if multiple pages were visible at once.

### DIFF
--- a/src/PdfReader/index.tsx
+++ b/src/PdfReader/index.tsx
@@ -4,6 +4,7 @@ import { Document, PageProps, pdfjs } from 'react-pdf';
 import {
   DEFAULT_HEIGHT,
   DEFAULT_SHOULD_GROW_WHEN_SCROLLING,
+  IN_VIEW_DELAY_MS,
   MAIN_CONTENT_ID,
   READER_MARGIN,
 } from '../constants';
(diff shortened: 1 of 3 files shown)


src/PdfReader/index.tsx
```diff
@@ -219,28 +218,58 @@ export default function usePdfReader(args: PdfReaderArguments): ReaderReturn {
...
+      });
+    }
   }, [state.pageNumber, state.settings?.isScrolling, state.rendered]);
 
+  const scrollState = React.useRef({
+    ratios: new Map<number, number>(),
+    lastVisiblePage: state.pageNumber,
+    isInViewUpdate: false,
+    lastProgrammaticNavAt: 0,
+  });
```
alea12: Could we move `scrollState` declaration above the `useEffect` that references it since `useRef` calls aren't hoisted? Or was this ordering intentional?

jackiequach: Moved it up.

alea12 (approved):

PR #199, opened 2026-04-06
SCHOL-205: Upgrade packages and pin versions

jackiequach:
- Upgrades and pins chakra-ui and associated package versions
- Upgrades React to 18.3.1


alea12 (approved):
Looks good!

PR #201, opened 2026-04-23
SCHOL-175: Web reader AQA 

jackiequach:
Resolves AQA items 13-20 in the [items page QA doc](https://docs.google.com/document/d/1eT_J0WxwGtYVLlyMwcNcoCRB93mliZH0zq2VMKS3Kfk/edit?usp=sharing)

- To resolve item 20, `Tabs` component was removed and replaced with `ToggleGroup` (radio group)


src/ui/Header.tsx
```diff
@@ -81,15 +81,28 @@ export default function Header(
...
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') {
+        handleFullscreen();
+      }
+    };
+    document.addEventListener('keydown', handleKeyDown);
+    return () => {
+      document.removeEventListener('keydown', handleKeyDown);
+    };
+  }, [handleFullscreen, isFullscreen]);
```
alea12: I'm not sure if `useEffect` would be necessary here. Are there any reasons we aren't using `isFullScreen` from the custom hook?

```ts
const [, toggleFullscreenHook] = useFullscreen();
```

jackiequach: When `toggleFullScreen` is passed to the reader from a consuming app, there is custom fullscreen behavior implemented with it. The `isFullScreen` from the hook only checks for `fullscreenchange` events on the browser.  The `useEffect` is needed to add a listener to call handleFullscreen to exit the app-controlled fullscreen state. 

With this, I found that we do need `isFullScreen` from the hook to properly handle the aria-labels and icon changes. This will be updated in my next commit.

alea12 (approved):
Thanks @jackiequach, looking good!

PR #204, opened 2026-05-13
Support navigation based on `start` URL parameter

alea12:
Refactored the reducer to handle the `?start` URL parameter on load. The logic was moved from a utility function edge case in `goToLocation()` to the `PDF_PARSED` action for better consistency

### DIFF
--- a/src/PdfReader/reducer.ts
+++ b/src/PdfReader/reducer.ts
@@ -155,7 +155,7 @@ export function makePdfReducer(
         const { pageNumber: currentPage, resourceIndex } = state;
 
         const currentHref = manifest.readingOrder[resourceIndex]?.href;
-        const startPage = getStartPageFromHref(currentHref) ?? 0;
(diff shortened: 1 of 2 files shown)


src/PdfReader/reducer.ts
```diff
@@ -161,15 +150,24 @@ export function makePdfReducer(
 
       // called when the resource has been parsed by react-pdf
       // and we know the number of pages
-      case 'PDF_PARSED':
+      case 'PDF_PARSED': {
+        const { numPages } = action;
+        const { pageNumber: currentPage, resourceIndex } = state;
+
+        const currentHref = manifest.readingOrder[resourceIndex]?.href;
+        const startPage = getStartPageFromHref(currentHref) ?? 0;
```
jackiequach: startPage should default to 1 to be consistent with the rest of this reducer.

jackiequach (approved):
looks good, just a minor comment

PR #207, opened 2026-08-11
SCHOL-857: New pdf.js reader

jackiequach:
[Confluence doc](https://newyorkpubliclibrary.atlassian.net/wiki/spaces/DIGTL/pages/1853390858/PDF+Reader)
[SCHOL-857](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-857)

This PR will not introduce any breaking changes to the existing web reader.

`PdfReader` is replaced with an implementation using pdf.js directly instead of react-pdf. 

## Initial load

Every page is always in the DOM as a fixed-size `div` placeholder whose dimensions are set from the page's intrinsic size (fetched via `pdfDoc.getPage(i)` at load time, before any canvas is painted). This gives the scroll container a correct, stable total height and a working scrollbar immediately, before any rendering has happened. The canvas, text layer, and annotation layer inside each page are only painted once the page scrolls into an `IntersectionObserver` window (300px above/below the viewport) and are torn down again when the page scrolls out.

## State management
Reader state (current page, total pages, scale, fit mode, rotation, navigation request counter) is managed by a `pdfReaderReducer` with a `useReducer`. Actions mirror the pattern in `usePdfReader`'s reducer but are simpler because it is a single-file viewer. 

**Note**: there is no multi-resource state machine (INACTIVE → FETCHING_RESOURCE → RENDERING_IFRAME → READY) implemented in the new reader to simplify the implementation.

Actions that require DOM measurements before the state change can commit (ZOOM_IN, ZOOM_OUT, ROTATE_CCW) are queued as a `pendingAction` state rather than dispatched directly. `PdfReader` processes them in a `useEffect`, calling `captureViewportAnchor()` synchronously against the current DOM before passing the action to the reducer. This preserves the pre-change scroll position across layout recalculations.

### Scroll position preservation (viewport anchor)
On every zoom or rotation, before the state change is committed, `captureViewportAnchor()` records:

- Which page sits at 30% down the viewport (`SCROLLSPY_ANCHOR_RATIO`)
- What fraction of that page's height the anchor point falls at (`intraPageRatio`)
- The pixel distance from the top of the viewport to the anchor (`viewportOffset`)

After the state change renders, a `useLayoutEffect` runs synchronously before the browser paints, recomputes the anchor page's new `top` and `height` from the updated scale/rotation, and sets `scrollTop` to restore the exact same content to the same screen position. The anchor is idempotent, a second capture within the same user action (e.g. rotate + refit scale) is a no-op, so both state writes resolve against the same pre-action snapshot.

## CSS over Chakra

Plain CSS is the better fit for the PDF reader since it is better for performance. Chakra's runtime style injection and prop-driven re-renders add overhead that compounds across all of the page components, whereas CSS rules are static and resolved entirely by the browser. The viewer also needs control over text/annotation layer positioning (position: absolute, transform-origin, z-index stacking) that maps naturally to CSS but would be awkward to express through Chakra's prop system.

## Testing
Locally replace the urls in `single-resource-short.json` with a larger pdf and go to `/pdf/single-resource-short` or `pdf/fixed-height-embedded-collection`

## TODO in separate PR:
- Remove `addTocToManifest` since the TOC is now retrieved directly in `PdfReader`. (breaking change, will need to be removed from consuming apps)
    - Remove `react-pdf` dependency 

[SCHOL-857]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/PdfReader/constants.ts
alea12, line 1: We have another constants configuration in `src/constants.ts`. Could you add a brief comment on what kind of constants should live where?

src/PdfReader/types.ts
```diff
@@ -1,68 +1,63 @@
...
+  pdfWorkerSrc?: string;
+  height?: string | number;
+  className?: string;
+  toggleFullScreen?: () => void;
+}
 
-export type ActiveState = ReaderState &
-  InternalState & { state: 'ACTIVE'; settings: ReaderSettings };
+export interface PdfReaderContentProps {
+  fileUrl: string | undefined;
```
alea12: This and `className` are never accessed in `index.tsx`. We should remove them if we're not implementing those

alea12: It would be great if you could do a similar scan to newly defined types!

src/PdfReader/PdfReaderContent.tsx
```diff
@@ -0,0 +1,544 @@
...
+    onOutlineLoad([]);
+    setPageBaseSizes([]);
+    setVisiblePages(new Set());
+    containerRefs.current.clear();
+
+    const loadDocument = async () => {
+      try {
+        if (!fileUrl) {
+          throw new Error('A PDF fileUrl is required');
+        }
```
alea12: Not a blocker; prop design suggestion - I think the boundary here should be such that this child component does not allow undefined `fileUrl`s. One way to tighten this would be to define `PdfReaderProps` to guarantee that either `webpubManifestUrl` or  `manifest` is present, so `usePdfReader` will always be able to resolve a `fileUrl`. Related, `resolveResourceUrl` should throw an error if `href` is not accessible.

src/PdfReader/index.tsx
```diff
@@ -1,536 +1,201 @@
...
-  const zoomOut = React.useCallback(async () => {
-    dispatch({
-      type: 'SET_SCALE',
-      scale: state.scale - SCALE_STEP,
-    });
-  }, [state.scale]);
-
-  const rotateCounterClockwise = React.useCallback(async () => {
-    dispatch({ type: 'ROTATE_COUNTER_CLOCKWISE' });
+  const [pdfLoadFailed, setPdfLoadFailed] = useState(false);
```
alea12: `pdfLoadFailed` is never set to `true`.  I think this would result in an infinite skeleton state without an indication that PDF fetch/render failed

src/PdfReader/reducer.ts
```diff
@@ -1,255 +1,118 @@
...
+  fitMode: FitMode;
+  rotation: number;
+  navigationRequestId: number;
+};
+
+export type PdfReaderAction =
+  | { type: 'GO_FORWARD' }
+  | { type: 'GO_BACKWARD' }
+  | { type: 'GO_TO_PAGE'; page: number }
+  | { type: 'GO_TO_HREF'; href: string }
```
alea12, line 17: These navigation actions hijack root browser's scroll position:

https://github.com/user-attachments/assets/d399ef3e-a03b-4541-9f6e-e9e8e52da83f


Is this a desired behavior?

alea12 (changes requested):
Thank you @jackiequach for this work! I appreciate you working on refactoring along the way.  I confirmed in-app experience like zoom & rotate has become way snappier.  Left some comments for your review

alea12:
Also, direct page navigation seems to have degraded (http://127.0.0.1:3000/item/1e21d527-35cf-45e3-8c51-f1e60e92fe5d?previewItemId=24077967&previewPage=00000015). Would this be something that we want to handle in the enhanced-search FE?

alea12:
For future reference, here are the steps I took to test on a separate repo:

<details>
<summary>Local dev build + testing with <code>enhanced-search</code></summary>

**1. Build & pack this branch**

```bash
cd web-reader
npm run build
rm -f nypl-web-reader-*.tgz  # clean up old builds if exists
npm pack                     # → nypl-web-reader-5.0.2.tgz
```

**2. Install the tarball into the enhanced-search web app**

```bash
cd ../enhanced-search/web
npm install ../../web-reader/nypl-web-reader-5.0.2.tgz
rm -rf .next
npm run dev
```

After testing, restore with: `git checkout package.json package-lock.json && npm install`.
</details>

src/PdfReader/constants.ts
jackiequach: Added comments in both files.

src/PdfReader/PdfReaderContent.tsx
jackiequach: Updated to resolve the `fileUrl` in `usePdfReader` instead and throw an error there.

src/PdfReader/index.tsx
jackiequach: Updated to properly set an error state when fetch/render fails. Added a per page render error using the existing error ui: 
image

src/PdfReader/reducer.ts
jackiequach: Fixed, it scrolls only the internal container.

jackiequach:
> Also, direct page navigation seems to have degraded (http://127.0.0.1:3000/item/1e21d527-35cf-45e3-8c51-f1e60e92fe5d?previewItemId=24077967&previewPage=00000015). Would this be something that we want to handle in the enhanced-search FE?

ES needs to be updated to use the fragment identifier `#page=X` since the PDF reader is now following these [PDF standards](https://pdfa.org/pdf-fragment-identifiers/).

We do need a separate to ticket to handle passing a page number to the web reader from the relevant snippet links so that the reader doesn't need to fully reload.

alea12:
@jackiequach pushed a very minor fix https://github.com/NYPL/web-reader/pull/207/commits/dd61c7d173ade170a4d0425a59031d47028eecc4. Otherwise looking great, thank you!

alea12 (approved):

PR #210, opened 2026-09-04
SCHOL-853: Enable range requests for linearized pdfs

jackiequach:
[SCHOL-853](https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-853)

## Description

Enables range requests for linearized pdfs to better view large documents.

- `getDocument()` now passes `disableStream: true`, `disableAutoFetch: true`, and `rangeChunkSize` to 1MB (from pdf.js's 64KB default). 

### Windowed page size measurement
- For non-linearized PDFs `PdfReaderContent` calls `pdfDoc.getPage(i)` for every page up front to measure layout height/width.
- For linearized PDFs, it seeds every page with a placeholder size (`DEFAULT_PAGE_SIZE`) so the scroll container exists right away, then only measures the real size of the current page and currently-visible pages (± `PAGE_MEASURE_BUFFER`) as the user scrolls or navigates.
- The placeholder for not-yet-measured pages is updated once the first batch is measured, using the *largest* width/height seen in that batch.
- Scroll position is preserved across placeholder to actual size swaps using the existing `captureViewportAnchor`.

### Lazy Table of Contents
- `resolveOutline()` calls `pdfDoc.getPageIndex()` once per outline entry, which can require a separate range request per entry for documents whose outline touches most of the file.
- Outline resolution is now deferred until the user requests it (opens the TOC panel).

### Loading UX (`PdfPage.tsx`)
- Added a loading indicator shown while a page's canvas/text/annotation layers are still rendering.

## Testing
- Manually verified against local 1230 page PDF initial loads:
    - Normal network (240 Mbps)
        - Before linearization: 834,226 KB (834.22 MB) transferred, 3.8 s
        - After linearization: 6,897 KB (6.897 MB) transferred, 813 ms
    - "Fast 4G" throttling
        - Before linearization: 834,226 KB transferred, 13.8 min
        - After linearization: 6,905 KB (6.905 MB) transferred, 13.39 s

## Future work
- Page size placeholders are not the best for Enhanced Search since the documents do not have uniform page dimensions. If largest page dimensions become available from the manifest then the logic around page measurement can be simplified.

[SCHOL-853]: https://newyorkpubliclibrary.atlassian.net/browse/SCHOL-853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/PdfReader/PdfReaderContent.tsx
+++ b/src/PdfReader/PdfReaderContent.tsx
@@ -562,7 +562,10 @@ const PdfReaderContent = ({
   );
 
   useEffect(() => {
-    if (fitMode) applyFitScale(fitMode);
+    if (fitMode) {
+      userZoomedRef.current = false;
+      applyFitScale(fitMode);
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [fitMode, pageBaseSizes.length]);
 
@@ -635,7 +638,9 @@ const PdfReaderContent = ({
         suppressNextResizeFitRef.current = false;
         return;
       }
-      if (fitModeRef.current) applyFitScaleRef.current(fitModeRef.current);
+      if (fitModeRef.current && !userZoomedRef.current) {
+        applyFitScaleRef.current(fitModeRef.current);
+      }
     });
     ro.observe(wrap);
     return () => ro.disconnect();
--- a/src/PdfReader/PdfReaderContent.tsx
+++ b/src/PdfReader/PdfReaderContent.tsx
@@ -12,7 +12,9 @@ import { MAIN_CONTENT_ID } from '../constants';
 import ReaderErrorAlert from '../ui/ReaderErrorAlert';
 import PdfPage from './PdfPage';
 import {
+  DEFAULT_PAGE_SIZE,
   PAGE_GAP,
+  PAGE_MEASURE_BUFFER,
   PAGE_PADDING,
   RENDER_ROOT_MARGIN,
   SCROLLSPY_ANCHOR_RATIO,
@@ -45,6 +47,7 @@ const PdfReaderContent = ({
   clearPendingAction,
   onOutlineLoad,
   onPageSizesReady,
+  loadOutline,
   onError,
 }: PdfReaderContentProps): React.ReactElement => {
   const [pdfDoc, setPdfDoc] = useState<PDFDocumentProxy | null>(null);
@@ -60,6 +63,7 @@ const PdfReaderContent = ({
   const [pageRenderErrors, setPageRenderErrors] = useState<Map<number, Error>>(
     new Map()
   );
+  const [rangeCapable, setRangeCapable] = useState<boolean>(false);
 
   const rootRef = useRef<HTMLDivElement | null>(null);
   const viewportWrapRef = useRef<HTMLDivElement | null>(null);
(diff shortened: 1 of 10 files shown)


src/PdfReader/pdfReader.css
```diff
@@ -81,6 +81,30 @@
...
+}
+
+.pdf-page-spinner {
+  width: 28px;
+  height: 28px;
+  border: 3px solid rgba(255, 255, 255, 0.2);
+  border-top-color: rgba(255, 255, 255, 0.7);
+  border-radius: 50%;
+  animation: pdf-page-spin 0.8s linear infinite;
+}
```
alea12, line 100: note: ask for design review

src/PdfReader/PdfReaderContent.tsx
```diff
@@ -338,6 +351,159 @@ const PdfReaderContent = ({
...
+        try {
+          const page = await pdfDoc.getPage(p);
+          if (cancelled) return;
+          const vp = page.getViewport({ scale: 1, rotation: 0 });
+          updates.push({
+            index: p - 1,
+            size: { width: vp.width, height: vp.height },
+          });
+          page.cleanup();
+          measuredPagesRef.current.add(p);
```
alea12: `measuredPagesRef.current.add(p)` runs before the sizes are actually applied. If the effect is cancelled mid-loop (e.g. `pageNumber` or `visiblePages` changes while scrolling), the `updates` collected so far are dropped, but those pages seem to stay marked as measured. They are then filtered out on the next run and never get their real size.

I would suggest moving the `add()` calls to right after `setPageBaseSizes(next)`, or removing the pending pages from the set on cancel. Similar issue exists in the full-measure loop below as well.

src/PdfReader/PdfReaderContent.tsx
```diff
@@ -60,6 +63,7 @@ const PdfReaderContent = ({
   const [pageRenderErrors, setPageRenderErrors] = useState<Map<number, Error>>(
     new Map()
   );
+  const [rangeCapable, setRangeCapable] = useState<boolean>(false);
```
alea12: Two questions here:

1. `rangeCapable` starts as `false`, and the probe races with `getDocument()`. If the document resolves first, could the "measure every page" effect for non-linearized PDFs start running on a linearized PDF? If that is the case, it would issue unnecessary range requests for all pages before it gets cancelled. Could we use `boolean | null` (null = unknown) and skip both measurement effects until the probe has finished?

2. The probe response is never aborted. If the server ignores `Range` and returns 200, will the browser keep downloading the full file body in the background? If so, that would mean we are downloading the PDF twice. We may want to abort the request after reading the headers (or `res.body?.cancel()`).

alea12 (changes requested):

src/PdfReader/PdfReaderContent.tsx
jackiequach: Updated to defer `measuredPagesRef` updates until after the cancellation checks

src/PdfReader/PdfReaderContent.tsx
jackiequach: 1. `rangeCapable` is now `boolean | null`. Both measurement effects check for the resolved value explicitly (rangeCapable !== true / rangeCapable !== false), so neither runs until the probe settles.
2. The probe fetch now uses an `AbortController` and calls `res.body?.cancel()` right after reading headers, so a server that ignores `Range` and returns a full 200 body won't have its body streamed in the background.

alea12 (approved):
Looking good, thanks @jackiequach!

